PR #6, opened 2017-08-05
Ensure compatibility with PHP version and REDCap version before enabling module

taylorr4:
Ensure compatibility with PHP version and REDCap version using config values when enabling a module. It will check the JSON attributes seen in the example below. If any attribute is missing or blank, it will be ignored during the check.

"compatibility": {
	"php-version-min": "5.4.0",
	"php-version-max": "",
	"redcap-version-min": "7.0.0",
	"redcap-version-max": ""
}

### DIFF
--- a/classes/ExternalModules.php
+++ b/classes/ExternalModules.php
@@ -994,6 +994,31 @@ class ExternalModules
 		require_once $path;
 	}
 
+	# Ensure compatibility with PHP version and REDCap version during module installation using config values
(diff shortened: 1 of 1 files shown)


mmcev106:
In it's current location, this code will run every time a module is used in any hook.  What do you think about moving this check to the enableForProject() method so the check is only performed at enable time?

taylorr4:
Thanks for catching that, @mmcev106 . How does this look now?

PR #9, opened 2017-08-07
More extensive logging + added confirmation prompt when disabling module

taylorr4:


manager/project.php
```diff
@@ -54,6 +54,25 @@
...
+				for the current project?
+			</div>
+			<div class="modal-footer">
+				<button data-dismiss="modal">Cancel</button>
+				<button id="external-modules-disable-button-confirmed" class="save">Disable module</button>
+			</div>
+		</div>
+	</div>
+</div>
+
```
mmcev106: It looks like this dialog is duplicated in control_center.php.  I think it would make sense to these to remove this code from project.php and control_center.php, and put it in templates/enabled-modules.php instead (which is already included in both of those files).

taylorr4:
Great idea! I was wondering about that, but wasn't sure where the best place would be to put it. How does this look now?

mmcev106:
Looks great!  Merged.

PR #10, opened 2017-08-07
Removed jQuery 3 + embedded JS/CSS from CDN

taylorr4:
Removed jQuery 3 since it was not being used. Also embedded the JS and CSS files that were previously being called via CDN.


manager/rich-text/get-uploaded-file-list.php
```diff
@@ -66,8 +66,6 @@
 	}
 </style>
 
-<script src="https://cdnjs.cloudflare.com/ajax/libs/jquery/3.2.1/jquery.min.js" integrity="sha256-hwg4gsxgFZhOsEEamdOYGBf13FyQuiTwlAQgxVSNgt4=" crossorigin="anonymous"></script>
-
 <button id="external-modules-rich-text-upload-button">Upload a file</button>
```
mmcev106: I just tested the Configuration Example module on this branch, and rich text support seems to be broken.  I suspect removing this line might have something to do with it.  I believe this file is loaded within an iframe in tinymce, so that is why jquery was required (though not necessarily jquery 3).  I don't know the best solution off the top of my head, but let me know if I can answer any questions.

taylorr4:
@mcguffk, thanks for pointing that out. I had messed up that function in a previous commit, but it was correct to begin with. Sorry about that.  I've reverted it here to remove the conflict.

@mmcev106, it turns out that the "jquery3 issue" wasn't really related to jquery3 being removed at all but due to assets that were apparently loaded automatically via the CDN for tinymce and select2. I've included those assets in the package now so that everything is bundled and included. So i can't see any issues right now when testing it while still having jquery3 *not* included. If all this tests fine and looks good, please go ahead an merge to testing. Thanks!

mmcev106:
I'm afraid we still need to either do some refactoring or include a local version of jquery on the **get-uploaded-file-list.php** page (since it loads within an iframe in tinymce).  I'm currently seeing a javascript error when trying to upload an image within a rich text field.

taylorr4:
Mark, this should work now. I've embedded jQuery 1 to replace jQuery 3, so it is no longer giving errors anymore with the tinymce iframe. Could you please check this? Thanks!

mmcev106:
Rich text image uploads seem to work great now.  Thanks!  Merging...

PR #12, opened 2017-08-07
Added REDCap's built-in field validation for text fields in a module's config file

taylorr4:
Added REDCap's built-in field validation as an option for text input fields set in a module's config file by using attribute "validation" with the additional option of using "validation_min" and/or "validation_max" as optional field-level attributes.

### DIFF
--- a/manager/css/style.css
+++ b/manager/css/style.css
@@ -51,6 +51,14 @@
 	width: 850px;
 }
 
+/* Modify Bootstrap modal z-index because it conflicts with jQuery UI modals */
+.modal-backdrop {
+	z-index: 99;
+}
+.modal {
+	z-index: 100;
+}
+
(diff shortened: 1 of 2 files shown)


manager/css/style.css
```diff
@@ -51,6 +51,14 @@
...
 
+/* Modify Bootstrap modal z-index because it conflicts with jQuery UI modals */
+.modal-backdrop {
+	z-index: 99;
+}
+.modal {
+	z-index: 100;
+}
+
 #external-modules-configure-modal input[type=radio] {
```
mmcev106: It looks like these lines are causing the REDCap menu bar to appear over the External Module dialogs and their backdrop.  The top edge of the dialogs are cut off slightly.

taylorr4: Okay, i will fix this. Thanks for catching it.

PR #14, opened 2017-08-07
update author information in module config

123andy:
* Added institution name to author config
* Added subject-helper to email link

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,7 +77,7 @@ External Modules combine the concepts of REDCap Plugins and REDCap Hooks. They c
 1. Open the **config.json** file. This is a JSON used to specify the External Module.
 	* Fill in the **name**
 	* Fill in the **description**
-	* Fill in the **name** and **email** for authors. At least one email is required to run the module.
+	* Fill in the **name** and **email** (and optional **institution**) for authors. At least one email is required to run the module.
 	* Grant **permissions** for all of the operations. Start with **hook_...**. For example, **hook_save_record**.
 	* **links** specify any links to show up on the left-hand toolbar. These include stand-alone webpages (substitutes for plugins) or links to outside websites. These are listable at the control-center level or at the project level.  A **link** consists of:
 		* A **name** to be displayed on the site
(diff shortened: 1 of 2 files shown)


taylorr4:
@123andy The mailto addition is a good idea. However, @mmcev106 has already merged in my code that adds the institution name (see in the testing branch). Note: In the future, i recommend you make changes based off the testing branch, and then submit pull requests to testing. There's a LOT of changes that Mark and I have made that are in testing but not in master yet. Making changes to testing is what we do now before Mark merges testing into master. Since these changes here are very small, I'll go ahead and merge this pull request into testing (hoping Mark doesn't mind).

123andy:
Thanks @taylorr4.

123andy:
@taylorr4 - Do you guys have a slack channel?  I've got a few random questions and wondering the best way to ask... For example, how does one enable a plugin globally for all projects (even if there is a pid in the url?

taylorr4:
@123andy, we do use Slack, but it's an internal Vanderbilt group. Perhaps we could think about a Slack group for External Modules. But for right now, i'd say you can ask the questions to me since i'm more familiar with ext mods now, and if i don't know, then i can ask @mcguffk or @mmcev106. I know those guys stay super busy, so i want to respect their time.

But regarding your question about enabling a plugin-type module page globally, i'm pretty sure that kind of feature has not been added yet (i could be wrong though). Our Data Core Team typically writes project-level plugins because they are typically doing customizations for individual groups of users, so i imagine they've likely only focused on project-level plugin functionality.

mmcev106:
Thanks for all the quick feedback Andy!  You can currently enable a module for all projects by default (and disable on a per project basis) by going to **Control Center -> External Modules**, clicking **Configure** for a given module, checking "Enable on all projects by default", then saving.

Rob, feel free to use your discretion on what can be committed directly and what might benefit from review.  Like you said, Kyle & I have other priorities, so we'll have to start passing the torch for External Modules to get the attention they deserve.

mmcev106:
Rob just stopped by and pointed out that you may have been asking about how to include a link to a module page that is not project specific.  You can currently add links in the Control Center via config.json.  This module provides an example:

https://github.com/mmcev106/redcap-external-module-example/blob/master/config.json

If you want to add your own custom link somewhere (perhaps inserted via javascript in a hook), you can get the url to a module page by calling $moduleClass->getUrl().  Hopefully that is helpful.  Don't hesitate to ask any more questions.

PR #17, opened 2017-08-10
add label to indicate global modules

123andy:
It is useful to know if a module is 'global' or 'per project optional'
at a glance


taylorr4:
@123andy, it sounds like a useful feature to me. I tested it, and it looks fine.

Issue #19, opened 2017-08-16
guidelines for external module development


wibeasley:
Start a document based on the conversation between @mcguffk and @123andy on Monday.  These guidelines are trying to anticipate the growth and lifecycle of modules.

pbchase:
@wibeasley, I really like this document. Thanks for the contribution.

mmcev106:
Closing, since the pull request was merged.

PR #20, opened 2017-08-16
start guidelines

wibeasley:
See #19


mmcev106:
Thanks Will!  I'm very appreciative of the time spent thinking about documentation.  It's not something I've had the luxury to do.  @mcguffk, I might leave the the ball in your court to review this since you were part of that discussion.  @taylorr4, since you'll likely be supporting whatever standards we create, I'm guessing you might want to weigh in as well.

PR #24, opened 2017-08-29
Add repeat_instance hook function parameter

lsgs:
From 7.3.4 REDCap includes a repeat instance parameter to the redcap_data_entry_form, redcap_data_entry_form_top, redcap_save_record, redcap_survey_complete, redcap_survey_page, and redcap_survey_page_top hook functions


mmcev106:
REDCap actually calls External Modules internally as of 7.6.1, and this hooks file has been removed in the **testing** branch.  Going forward, any hook parameter changes should automatically be supported in this codebase  (each module will still require changes).  I just added a note at the top of the README to encourage future development against the **testing** branch (it's always the latest).

PR #25, opened 2017-08-31
Andy

123andy:
minor patches


manager/templates/enabled-modules.php
```diff
@@ -185,7 +185,7 @@
 			?>
 				<tr data-module='<?= $prefix ?>' data-version='<?= $version ?>'>
 					<td><div class='external-modules-title'><?= $config['name'] . ' - ' . $version ?>
-                            <?php if ($system_enabled) print "<span class='label label-warning'>Enabled for All Projects</span>" ?>
+                            <?php if ($system_enabled) print "<span class='label label-warning'>Global Module</span>" ?>
```
mmcev106: We've intentionally moved away from the term "global" in favor of "system" or "system-wide".  Also, I'm a little concerned that "System-Wide Module" might not be meaningful enough to be obvious to most users.  It might be a little ambiguous since all modules are installed, enabled, and can be configured at the system level (even if they're not enabled on all projects).

123andy: I'm fine with whatever terminology you want to use - but there was certainly confusion - even among the presenters - on how to differentiate between 'installing' the code on your server, then 'enabling' a module on your server, then 'enabling' a module for all projects or certain projects on your server... 

What if in the control center - the modules box showed a label with the number of projects enabled (or ALL if the system-wide checkbox was checked)... That might help a little to differentiate.

mmcev106: I like the sound of a project count or "ALL".  It could replace the "View Usage" button.  I suppose "ALL" wouldn't be entirely accurate if it had been disabled on a couple of projects, but that might not matter.  There is definitely room for improvement.

taylorr4: I think the term "system" probably makes more sense/is more clear than "global". And i definitely like the idea of a project count or "All" there in the Control Center list of modules.

PR #31, opened 2017-09-25
Adding cron job functionality for modules

taylorr4:
If any cron jobs are defined in the config.json when a module is enabled, those jobs will get added as new rows in REDCap's redcap_crons table. The cron_frequency and cron_max_run_time must be defined as integers (in units of seconds), and all attributes must be provided, else an exception is thrown, thus preventing the module from being enabled. 

The cron_max_run_time refers to the maximum time that the cron job is expected to run (once that time is passed, if the cron is still listed as processing, it assumes it has failed/crashed and thus will automatically enable it to run again at the next scheduled interval). The "method" attribute corresponds to a method in the module class. 

**Note: This External Module cron job functionality will only function in REDCap 7.6.6 and higher.**

Example config format for module crons. 
```
{
	"name": "Name of the module",
	...
	"crons": [
		{
			"cron_name": "cron1",
			"cron_description": "Cron that runs every 30 minutes to do X",
			"method": "cron1",
			"cron_frequency": "1800",
			"cron_max_run_time": "60"
		 },
		{
			"cron_name": "cron2",
			"cron_description": "Cron that runs daily to do Y",
			"method": "some_other_method",
			"cron_frequency": "86400",
			"cron_max_run_time": "1200"
		 }
	]
}

```

### DIFF
--- a/classes/ExternalModules.php
+++ b/classes/ExternalModules.php
@@ -371,6 +371,9 @@ class ExternalModules
 	static function disable($moduleDirectoryPrefix)
 	{
 		self::removeSystemSetting($moduleDirectoryPrefix, self::KEY_VERSION);
+		// Disable any cron jobs in the crons table
+		$instance = self::getModuleInstance($moduleDirectoryPrefix, $version);
+		self::removeCronJobs($instance);
 	}
 
 	# enables a module system-wide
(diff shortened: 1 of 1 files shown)


123andy:
Awesome!

mmcev106:
Very cool!!!  I can't wait to give it a whirl.

mmcev106:
This request has been merged!   Thanks very much.  Per chance, is it worth your time to add an example cron to the configuration example module and/or the README documentation?

taylorr4:
@mmcev106, yes, indeed. I'll work on that. Thanks!

123andy:
If a module is deleted, is the corresponding cron entry in the crons table cleaned up somehow?  What about if the module is disabled?  Will that prevent the cron from firing or are we relying on the EM code to do that?

taylorr4:
If a module that was downloaded from the REDCap Repo (central repository) is disabled via the UI, then the crons get deleted from the crons table, and if it gets re-enabled, the crons get re-added. But if the module directory was deleted via the backend, then the crons would remain and be orphaned. One of the few things left on my to-do list is to build a new REDCap cron job that validates module crons to make sure they are in still sync, either for the reason of backend deletion OR in case someone modifies the config via backend methods on the server *after* the module has been enabled. So this new cron would keep all the module crons in sync to prevent issues like these. So i feel comfortable that we're covering our bases here.

123andy:
awesome

PR #32, opened 2017-09-26
Adds "delete module" functionality in the UI

taylorr4:
Adds a "delete module" button in the Available Modules dialog in the Control Center. When clicked (after a confirmation), it will remove the module directory from the REDCap server. Note: Example modules (those included in the Ext Mod package) are not allowed to be deleted, nor are modules that are enabled. A module must be disabled before it can be deleted.

Screenshot:
delete-module

### DIFF
--- a/classes/ExternalModules.php
+++ b/classes/ExternalModules.php
@@ -1722,14 +1722,29 @@ class ExternalModules
 
 	# formats directory name from $prefix and $version
 	static function getModuleDirectoryPath($prefix, $version){
+		// Look in the main modules dir and the example modules dir
 		$directoryToFind = $prefix . '_' . $version;
 		foreach(self::$MODULES_PATH as $pathDir) {
 			$modulePath = $pathDir . $directoryToFind;
 			if(is_dir($modulePath)) {
-				break;
+				return $modulePath;
 			}
 		}
-		return $modulePath;
+		// If module could not be found, it may be due to load balancing, so check if it was downloaded
+		// from the central ext mod repository, and redownload it
+		if (!defined("REPO_EXT_MOD_DOWNLOAD") && self::wasModuleDownloadedFromRepo($directoryToFind)) {
+			$moduleId = self::getRepoModuleId($directoryToFind);
+			if ($moduleId !== false) {
+				// Download the module from the repo
+				$status = self::downloadModule($moduleId, true);
+				if (!is_numeric($status)) {
+					// Return the modules directory path
+					return dirname(APP_PATH_DOCROOT).DS.'modules'.DS.$directoryToFind;
+				}
+			}
+		}		
+		// Still could not find it, so return false
+		return false;
 	}
 
 	static function getModuleDirectoryUrl($prefix, $version)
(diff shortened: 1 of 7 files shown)


taylorr4:
Note that in this pull request i also snuck in the logic to deal with the load balancing issue.  Sorry to stick in several things at once, but much of it was dependent on each other.

mmcev106:
No prob!  I'm looking forward to skimming through it (likely later this week).

123andy:
Should there be an option to 'enable auto-downloading of missing modules'.  I would hate for someone to delete code off their server and then see it keep reappearing.  That could be pretty annoying.  Perhaps if the code is auto-downloaded there should also be a super-user email warning that is issued that explains what happened?

taylorr4:
@123andy, i honestly think there should _not_ be an option for auto-downloading. If one were added, i think most people would be confused by what it does and why it's there. I think it would only stand to complicate things. The way i see it is that most places will download modules via the UI from our repo. There will be a minority who side-load them (place them directly on the server via back-end) like developers. The auto-downloading will only apply for those modules downloaded from our repo, and there's a new db table to track that. I think it would be pretty unlikely that someone would download a module via the UI and then delete it via the back-end. Those are the people who would be adversely affected by the auto-downloading. And in that case, i think having it auto-download far outweighs the downside of having the module not be there when it is needed during a load balancing situation, which would be very hard to explain to someone how to duplicate the code they downloaded onto their other server, especially if they are not technical. So having auto-download as a possible option in my opinion only stands to open up the possibly for these issues, while the downside is merely that a folder won't stay gone. And knowing what i know about REDCap admins, most people that will be downloading modules via the UI are only half likely to even know that their server is load balanced, if it is.

So that's my two cents. Ultimately i think there should just be some education about deleting modules via the UI if you downloaded via the UI.

123andy:
Hey @taylorr4 - as usual - I'm not communicating very clearly...   I didn't mean to suggest blanket auto-downloading.   I agree with your logic above!   

I was just wondering if there should be an option to 'enable/disable' load balancing behavior so any features like this would be more clear?   Just opening the discussion.  I can live with it being default.

One other use case - If someone has more than one web server and they delete a module, it is going to remove the module folder from the active server but currently will leave the orphaned copy on the other nodes, right?  You could imagine a cron that works to replicate additions and deletions of modules across nodes...  FWIW - we're planning to move to a multi-node frontend in our next release.  Similar to Dmitry at Partners, we're excited to isolate the API calls from our main web server...

taylorr4:
@123andy  yes, i don't necessarily disagree with you, but i'm not able to imagine in my mind how the UI would be and how people would understand it. So i think it should be enabled by default to assume that a server has load balancing (just in case the REDCap admin doesn't know), but it would be confusing if they the admin saw that option and wondered why REDCap was assuming they had load balancing. Plus there's the question of where the setting would go. If we put it on the User Settings page in Control Center, it means no one would likely ever see it or realize it was connected to how Ext Mods work (why include it if not easily findable?). And it probably wouldn't make sense to put it on the Modules Manager page because then i think it would be too prominent since there aren't any other options. I just can't think of a nice implementation for that setting is the issue i'm having with it.

Regarding deleting a module with load balancing, that is a tough one. I had mistakenly forgotten about that scenario. Let me think more about that one. I think i can modify my new redcap db table to log if a module was deleted and thus use that to signify that a directory should be deleted if we somehow see it again on another node of the server.

PR #34, opened 2017-09-28
Fix isLocalhost() dependency on IP assumption

taylorr4:
Added usage of $is_development_server global setting in isLocalhost() to prevent the assumption that servers with an IP web address are always a dev/test/staging server. Note: $is_development_server exists only in REDCa 7.6.7 and later.


mmcev106:
Thanks Rob!  I think this might only affect Scott on a daily basis, but I might wait to merge this until 7.6.7 is available for download.

PR #36, opened 2017-10-02
Add user permission to modify project config

taylorr4:
Adding more granular controls for who can modify a module's project-level configuration settings. When enabling a module in the Control Center, REDCap admins can set a module to require module-specific permission (via a project's User Rights page) in order to allow them to modify its configuration in a project, else it reverts to the default, which is that it requires Project Setup/Design privileges to configure the module.
perm2

On User Rights page:
perm1

### DIFF
--- a/classes/ExternalModules.php
+++ b/classes/ExternalModules.php
@@ -41,6 +41,7 @@ class ExternalModules
 	const SYSTEM_SETTING_PROJECT_ID = 'NULL';
 	const KEY_VERSION = 'version';
 	const KEY_ENABLED = 'enabled';
+	const KEY_CONFIG_USER_PERMISSION = 'config-require-user-permission';
 
 	const TEST_MODULE_PREFIX = 'UNIT-TESTING-PREFIX';
 
@@ -98,10 +99,21 @@ class ExternalModules
 		),
 		array(
 			'key' => self::KEY_ENABLED,
-			'name' => 'Enable on all projects by default',
+			'name' => '<b>Enable module on all projects by default:</b><br>Unchecked (default) = Module must be enabled in each project individually',
 			'project-name' => 'Enable on this project',
 			'type' => 'checkbox',
 			'allow-project-overrides' => true,
+		),
+		array(
+			'key' => self::KEY_CONFIG_USER_PERMISSION,
+			'name' => '<b>Module configuration permissions in projects:</b><br>By default, users with Project Setup/Design privileges can '
+					. 'modify this module\'s project-level configuration settings. Alternatively, project users can be given explicit '
+					. 'module-level permission (via User Rights page) in order to do so',
+			'type' => 'dropdown',
+			"choices" => array(
+				array("value" => "", "name" => "Require Project Setup/Design privilege"),
+				array("value" => "true", "name" => "Require module-specific user privilege")
+			)
 		)
 	);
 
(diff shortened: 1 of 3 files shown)


taylorr4:
NOTE: The integration for these changes in REDCap (i.e., on User Rights page) will go into REDCap 7.6.8.

123andy:
Good work – I think all of these permissions are going to be important to make this a success.

Have we added permissions for links yet?  I’m also thinking the UI of the EM configuration page may need to be enhanced to more clearly separate these ‘global’ configs from project-specific or system-specific configs.  Any thoughts to converting it from a modal dialog to a full page?  I’d love to add to the config a single javascript method that gets called so you can perform custom cleanup on the config form.  As usual – ideas and no time.

taylorr4:
@123andy, i'm not sure what you mean by "permissions for links". What kind of links? 

I agree that perhaps we could make a visual distinction between the default global configs and module configs, but i don't see it as being very urgent yet.

Regarding converting the config from a modal to a full page, i think that's a decent idea, but i don't think it's very crucial at all right now + it would take a good amount of work to redo it. Yes, i think having a search feature and other things on that page would be good, but i think they can wait a little while. This thing doesn't have to be perfect on day 1.

I would really like to get External Modules integrated into REDCap **by the end of this month**, so anything that's not crucially needed by either users or developers can wait, in my opinion. EM will get iteratively better and evolve, just like REDCap does, so i recommend we put full effort on the "absolute must haves" right now.

**What other features/additions are still missing from EM that you all think should be required before we officially release it?** (this question is to anyone, btw)

scottjpearson:
Would anyone have trouble if I added three methods to the AbstractExternalModule class?

Background: I’m trying to integrate a plugin with an elaborate directory structure into an external module. The getUrl function is failing to deal with the directory structure.

Mark and I proposed adding the page variable at the end of the GET sequence from getUrl. Then, a prefix can be added to the directory structure.

Modified: getUrl
Add: getUrlBaseForModule
Add: getUrlGetParams
Add: reformatUrlGetParams

Anyone with objections or modifications?

Scott

biggeeves:
I like the 1st one.  getUrlGetParams is the idea to get the folder structure?  What would reformatUrlGetParams do?

taylorr4:
Hi @scottjpearson, sorry, but i'm a little confused. Doesn't getUrl already return &page=XXXX in the URL it returns? Also, would you mind briefly describing what each of those new methods do? I'm sure they're useful and good, but just curious, so that i understand them. Thanks!

mmcev106:
I'm not sure if Scott is monitoring these comments, but I don't think this message was intended for this issue.   We did discuss options internally and are leaning toward to simply making **getUrl()** return a base url when no parameters are passed.

123andy:
Re: permissions – this is a nice-to-have addition – not a must-have.

Let’s say you define a project-level link for a module but want to have user-level control over who sees the link.  One method is to apply some javascript/css via a hook but it is a lot of complexity.

I'm not sure what the best format is, maybe:

require-xxx (require-design / require-export)?


"links": {
   "project": [
      {
         "name": "My Module ",
         "icon": "gear",
         "url": "project_level_homepage.php",
         "require-design":true,                                           // this link would only be displayed in the left sidebar
//  if the current user had ‘design’ rights.
      }
   ]
},



Andrew Martin, PhD
Stanford University School of Medicine

taylorr4:
@123andy, i think it would definitely be useful in the future to have ways to set permissions for viewing the links, but i'm not quite sure what the ideal format would be for setting the rights. I think your suggestion is okay, but i'm not 100% in love with it. I can't think of anything any better right now though.

123andy:
Permission logic parsing :-) [design] = 1

PR #38, opened 2017-10-03
Remove author listing from Project Module Manager page

taylorr4:
Removes the author's name, email, and institution from the Project Module Manager (but still displays them on the Control Center Module Manager). @mmcev106 and I talked about this, and we both agreed that it would not be the best idea to list the author's name/email to end-users on that page. I would fear that end-users might suddenly start contacting the author, when it would instead be more appropriate for the REDCap administrator to contact the author (if they ever needed to). Plus i'm sure some authors might not want their name and email advertised to tens of thousands of REDCap users all around the world. This way, the admin remains the point person for the module and its author since the author info is still displayed on the Control Center Module Manager.

### DIFF
--- a/manager/ajax/get-disabled-modules.php
+++ b/manager/ajax/get-disabled-modules.php
@@ -69,6 +69,7 @@ require_once '../../classes/ExternalModules.php';
                         </div>
                         <div class='external-modules-byline'>
 <?php
+if (SUPER_USER && !isset($_GET['pid'])) {
 	if ($config['authors']) {
 		$names = array();
 		foreach ($config['authors'] as $author) {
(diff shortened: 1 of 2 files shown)


taylorr4:
@scottjpearson, as mentioned above, the author name/email is still listed on the control center page for the admins to see. The admins should be reporting bugs, if any occur. I don't think that end-users should be responsible for doing that. So given that, we don't see any reason to display the name/email of the author(s) to end-users.

taylorr4:
Oh, man, you're killing it today, @mmcev106! Thanks!

mmcev106:
Heh, gotta have a catch up day every now and then...

PR #40, opened 2017-10-06
sql command is no longer required

tranpl:
the "create tables.sql" file already contains the type column. I was following the direction on installing the external modules on this repo. After removing the first line "ALTER TABLE `redcap_external_module_settings` ADD COLUMN `type` VARCHAR(12) NOT NULL DEFAULT 'string' AFTER `key`;" on the migration-2017-01-18_10-03-00.sql , I was able to continue.


tranpl:
creating a new pull against testing branch...

PR #47, opened 2017-10-18
Support for "redcap_{hookname}" naming convention

taylorr4:
Instead of using "hook_{hookname}" as a module method, it now also alternatively supports "redcap_{hookname}", which seems more intuitive. After talking with Mark about this, i'll also cater the official ExtMods documentation to prefer the "redcap_{hookname}" convention.


123andy:
Great - I think this is a common point of confusion.

PR #49, opened 2017-10-23
Add module search feature on module manager page

taylorr4:
Adding an input text field above enabled/disabled module lists to allow quick searching of modules by title, author, description, etc.


taylorr4:
This is a pretty minimal and benign feature, so going ahead and merging.

PR #50, opened 2017-10-23
Keep module crons in sync in redcap_crons

taylorr4:
Adding new methods that will be run by a REDCap cron job in REDCap 7.6.10+ to ensure that a module's cron jobs (if a module has any crons defined in config.json) are always in sync so that their values in the redcap_crons table match the values on config.json. Note: The only real way they could get out of sync is if developers were modifying the config.json file directly on the server *after* a module has been enabled, but this will provide extra protection for that edge case.

### DIFF
--- a/classes/ExternalModules.php
+++ b/classes/ExternalModules.php
@@ -428,27 +428,36 @@ class ExternalModules
 		// Parse config to get cron info
 		$config = $moduleInstance->getConfig();
 		if (!isset($config['crons'])) return;
-		$externalModuleId = self::getIdForPrefix($moduleInstance->PREFIX);
 		// Loop through all defined crons
 		foreach ($config['crons'] as $cron) 
 		{
 			// Make sure we have what we need
-			self::validateCronAttributes($cron);
-			// Add the module
-			$sql = "insert into redcap_crons (cron_name, external_module_id, cron_description, cron_frequency, cron_max_run_time) values
-					('".db_escape($cron['cron_name'])."', $externalModuleId, '".db_escape($cron['cron_description'])."', 
-					'".db_escape($cron['cron_frequency'])."', '".db_escape($cron['cron_max_run_time'])."')";
-			if (!db_query($sql)) {
-				// If fails on one cron, then delete any added so far for this module
-				self::removeCronJobs($moduleInstance);
-				// Return error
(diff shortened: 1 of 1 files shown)


classes/ExternalModules.php
```diff
@@ -421,34 +421,43 @@ static function enableAndCatchExceptions($moduleDirectoryPrefix, $version)
 	}
 
 	# initializes any crons contained in the config, and adds them to the redcap_crons table
-	static function initializeCronJobs($moduleInstance)
+	static function initializeCronJobs($moduleInstance, $moduleDirectoryPrefix=null)
```
mmcev106: It looks like we're assuming the instance will exist.  We could pass $moduleInstance->PREFIX into the removeCronJob() call below and remove the $moduleDirectoryPrefix parameter to this method.

classes/ExternalModules.php
```diff
@@ -472,16 +481,106 @@ static function validateCronAttributes(&$cron=array())
...
 		}
+		// If method does not exist, then disable module
+		if (!empty($moduleInstance) && !method_exists($moduleInstance, $cron['method'])) {
+			throw new Exception("The external module \"{$moduleInstance->PREFIX}_{$moduleInstance->VERSION}\" has a cron job named \"{$cron['cron_name']}\" that is trying to call a method \"{$cron['method']}\", which does not exist in the module class.");
+		}
 	}
 
 	# remove all crons for a given module
-	static function removeCronJobs($moduleInstance)
+	static function removeCronJobs($moduleInstance, $moduleDirectoryPrefix=null)
```
mmcev106: For simplicity's sake, I'm wondering if it would be worth removing the $moduleInstance parameter, and only passing the prefix.  If the prefix isn't null, the instance can be retrieved using getModuleInstance().  Instances are cashed so there shouldn't be any performance hit.  Any existing calls could be changed to $instance->PREFIX.  What do you think?

taylorr4: That makes sense to me. I can go ahead and code these up, if you want (unless you have already).

mmcev106: I have not made any changes, so feel free if you think that's worthwile.  I'm happy merging this regardless.

taylorr4:
Okay, how does that look?

mmcev106:
Looks great to me!  Merged.

PR #51, opened 2017-10-25
Multiple module directories

123andy:
Added support for more than one module directory on a server for custom development.
For example, you may want to have a local module folder controlled by a local version control (e.g. modules_internal, or modules_staging)

The directories are controlled by a pipe-delimited array of paths stored in redcap_config.
// insert into redcap_config values ('external_module_alt_paths', '/modules_staging');

### DIFF
--- a/classes/ExternalModules.php
+++ b/classes/ExternalModules.php
@@ -207,8 +207,28 @@ class ExternalModules
 		} else {
 			$modulesDirectories = [dirname(APP_PATH_DOCROOT).DS.'modules'.DS, dirname(APP_PATH_DOCROOT).DS.'external_modules'.DS.'example_modules'.DS];
 		}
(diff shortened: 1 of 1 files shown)


taylorr4:
@123andy, this looks good and will be useful. Worked well when i tested it. I will add that setting to the config table in the next REDCap release (which will be 8.0.0). To allow admins to set the EM alt paths in the UI, I will add a new text box on the Control Center's Module Configuration page (which i now realize might be a little confusing since it has "Module" in the title, so i might add some text at the top of that page to clarify, or consider renaming the page).

@mmcev106, if you don't mind, i think i will go ahead and merge this into testing to help expedite things to prep for the 8.0 release and the opening of the REDCap Repo next week.

mmcev106:
Thanks Andy!

PR #60, opened 2017-11-02
Include "Text" in types list.

biggeeves:
Text was the last word of the line.  Updated to be first word in the unordered list of field types.


taylorr4:
@biggeeves, this change was already been incorporated into the testing branch earlier this week. Remember to always submit pull requests to testing rather than master. Thanks!

PR #63, opened 2017-11-07
Json objects for setSetting and getSetting

123andy:
Currently there is some odd behavior when trying to use ->saveSettings() with an object.  For starters, calling the method twice will cause a fatal error since the oldValue comparison isn't correct. 
 
I see that arrays are being stored as json, but I think both arrays and objects should be stored as json to more flexible handle parameters.  

I am suggesting the existing type be changed from 'json' to 'json-array' or 'json-object'.
Ideally, this patch would include a sql statement to disambiguate between json and json-array but in the meantime it works with both.

Consider adding:
`update redcap_external_modules_settings set type='json-array' where type='json'`


taylorr4:
@123andy, I don't have time to thoroughly test this soon or evaluate what this might affect, but it seems fine from what i can tell. I'll leave it up to @mmcev106 's discretion to review this, and if all is good, i can add that SQL update into the next version of REDCap.

mmcev106:
It also might take me a while before considering this reaches the top of my TODO list.  My initial thought is that there might be a way to fix the scenarios you encountered without the added complexity of splitting the data type, but I'm guessing you may have already hit a dead end there.

123andy:
If you try to save an object, it gets written as type object without any content in the current iteration so it is a potential bug that will cause people problems.  

My first thought was to just push objects as json as well, but since json currently implies array (with the json-decode passing the assoc flag) this doesn't work.

So, option 1 is to just add "json-object' and leave the existing "json" to represent json-encoded arrays.  This fixes the object bug, but to make it cleaner when viewing the table I also changed "json" to "json-array".  In looking at the code, the only type the type is manually set is for files, so I don't think this should break anything.

mmcev106:
One option I was considering was perhaps just detecting whether the first character was a '[' or a '{'.

123andy:
I think a better option, potentially, is to mark the datatype as 'array' and just by convention always store arrays or objects as json.  Then the type would mean what the object in the table really is.  It would be convention that 'array', 'object', and 'boolean' are json-decoded and encoded automatically whereas 'string' is not.

123andy:
Or, back to option 1 - just call it 'object' and imply that it is json-encdoed/decoded automatically...

mmcev106:
I looked into this a little bit and tried to write a unit test, but I am having trouble identifying a scenario where this change makes a difference.  It looks like json_decode() automatically handles detection of objects vs. arrays.  Prior to this change, I am not seeing a scenario where getSetting() does not return exactly what is passed into setSetting() for PHP Arrays and basic types.  What scenario am I missing?  At the moment I'm wondering if part of the problem might be something specific to saveSettings().

mmcev106:
I just [expanding the setting type checking unit test a bit](https://github.com/vanderbilt/redcap-external-modules/commit/5fdf7754c46025e23a729666047c6d6e20796c8f), in case anyone is curious.

123andy:
The first time you run this:
`
<?php
$arr = array("Foo" => "Bar");

$obj = new StdClass();
$obj->foo = "bar";

$module->setProjectSetting("arr", $arr);
$module->setProjectSetting( "obj", $obj);

$arr_out = $module->getProjectSetting("arr");
$obj_out = $module->getProjectSetting("obj");

print "<pre>";
print "\narr: " . print_r($arr,true);
print "\narr out: " . print_r($arr_out,true);
print "\nobj: " . print_r($obj,true);
print "\nobj out: " . print_r($obj_out,true);

exit();
`

You get the following output:
`
arr: Array
(
    [Foo] => Bar
)

arr out: Array
(
    [Foo] => Bar
)

obj: stdClass Object
(
    [foo] => bar
)

obj out: stdClass Object
(
    [scalar] => 
)
`

and the following in the database:
`
18,arr,json,{"Foo":"Bar"}
18,obj,object,
`

Now, if you refresh the script again and try to overwrite obj again, you get an error.
`
Fatal error: Uncaught Exception: Unexpected number of affected rows (0) on External Module setting query: UPDATE redcap_external_module_settings SET value = '', type = 'object' WHERE external_module_id = 4 AND project_id = '18' AND `key` = 'obj' in /Users/andy123/Projects/redcap-local/external_modules/classes/ExternalModules.php:819 Stack trace: #0 /Users/andy123/Projects/redcap-local/external_modules/classes/ExternalModules.php(643): ExternalModules\ExternalModules::setSetting('shazam', '18', 'obj', NULL) #1 /Users/andy123/Projects/redcap-local/external_modules/classes/AbstractExternalModule.php(199): ExternalModules\ExternalModules::setProjectSetting('shazam', 18, 'obj', Object(stdClass)) #2 /Users/andy123/Projects/redcap-local/modules/shazam_v0.1/config.php(10): ExternalModules\AbstractExternalModule->setProjectSetting('obj', Object(stdClass)) #3 /Users/andy123/Projects/redcap-local/external_modules/index.php(54): require_once('/Users/andy123/...') #4 {main} thrown in /Users/andy123/Projects/redcap-local/external_modules/classes/ExternalModules.php on line 819
`

While I don't think the actual config.json editor is affected, if people are using the abstract getSetting/setSetting methods they could easily run into this scenario.

mmcev106:
I apologize, I thought we were talking about JSON objects, not PHP objects.  I'll look into this again as soon as I can.

mmcev106:
I just added a unit test using a basic stdClass example.  In general this change looks great.  It looks like moving the `$type == "boolean"` if statement up a few lines breaks an edge case around updating existing boolean values.  Moving it back down seems to fix the issue, but I figured I should ask before committing that.  Was there a specific reason it was moved up?

123andy:
No reason..  Just grouping what I thought were similar blocks of code.  Sorry for messing up the other use case.  My changes really didn't have anything to do with boolean.

mmcev106:
No worries, there's a lot going on in that method...  Merged!  @taylorr4, I'll let you know when this moves to production so we can add the  `update` statement per Andy's suggestion.

mmcev106:
@taylorr4, this change is in production.  The update SQL Andy suggested is not required, but would normalize values.  Would it be possible to add that to an upcoming release?

taylorr4:
@mmcev106, thanks, will add this and the SQL query into REDCap 8.1.0.

Issue #13, opened 2017-08-07
Thoughts around enabling modules


123andy:
Currently when you go to enable a module in your server, the external modules presents a modal frame...  I'm thinking that over time there are going to be a LOT of modules and this interface is going to be less than ideal.

What about making it a normal page with a datatable.  This would make for easy searching, etc. Also, we could use the 'expand row' feature ([Data Table Rows](https://datatables.net/examples/api/row_details.html)) to make it easy to view more details on how the module works, etc...

Thoughts?

taylorr4:
@123andy it sounds like a good idea. Definitely will be useful inside projects where more modules will be displayed in the modal.

tbembersimeao:
I think it would be nice to have additional hooks like:
- `hook__module_enable`: at the moment the module is enabled (useful when some module depends on db changes to work, e.g. creating tables)
- `hook_module_disable` at the moment that the module is disabled (useful for garbage cleaning, e.g. removing tables)

mmcev106:
I think adding module event hooks would be very useful.  At the moment I'm envisioning the following:
- hook_module_system_enable
- hook_module_system_disable
- hook_module_system_change_version
- hook_module_project_enable
- hook_module_project_disable

tbembersimeao:
That sounds great!

123andy:
The idea was floated of being able to add a module-configurable 'validation' page.  Say, for example, a module requested that a new table be added to the database.  When the module is enabled, it could check if that table exists - and if it didn't, it could present the user with the sql for creating the table.

I also wonder about version confirmations - say someone wants to enable a new module - if the module had only been tested at version 7.6.2 and the current version was 7.6.6 the hook might be able to ask the user if they still want to proceed given that the module hadn't been tested against that version...

tbembersimeao:
@mmcev106 Created a PR to add the new hooks: https://github.com/vanderbilt/redcap-external-modules/pull/74.

pbchase:
We could really benefit from this feature. Today we have been designing yet another module that would benefit from it. In each case, the goal is to create a side-table in MySQL to serve the data-storage needs of the module. We want to make modules that are accessible to the non-technical REDCap community, but to do this we need more services from the External Modules framework. 

[...]

PR #74, opened 2017-12-18
Adding extra hooks to be triggered on external modules management operations

tbembersimeao:
Implements hooks described in the following issue: https://github.com/vanderbilt/redcap-external-modules/issues/13#issuecomment-323105463.

### DIFF
--- a/classes/ExternalModules.php
+++ b/classes/ExternalModules.php
@@ -824,7 +824,7 @@ class ExternalModules
 							'$value'
 						)";
 			} else {
-				if ($key == ExternalModules::KEY_ENABLED && $value == "false" && $pidString != "NULL") {
+				if ($key == self::KEY_ENABLED && $value == "false" && $pidString != "NULL") {
 					$version = self::getModuleVersionByPrefix($moduleDirectoryPrefix);
 					self::callHook('redcap_module_project_disable', array($moduleDirectoryPrefix, $version, $projectId));
 				}
--- a/classes/ExternalModules.php
+++ b/classes/ExternalModules.php
@@ -1179,6 +1179,26 @@ class ExternalModules
 		} else {
 			$hookName = substr(self::$hookBeingExecuted, 7);
 		}
+
+		$individualHooks = array(
+			'module_system_enable',
+			'module_system_disable',
+			'module_system_change_version',
+			'module_project_enable',
+			'module_project_disable',
+		);
+
+		if (in_array($hookName, $individualHooks)) {
+			// For individual hooks, the first argument is expected to be the
+			// module prefix.
+			$targetPrefix = array_shift($arguments);
+
+			// Preventing hook to be called from a non related module.
+			if ($targetPrefix != $prefix) {
+				return;
+			}
+		}
+
 		$hookNames = array('redcap_'.$hookName, 'hook_'.$hookName);
 		
 		if(!self::hasPermission($prefix, $version, 'redcap_'.$hookName) && !self::hasPermission($prefix, $version, 'hook_'.$hookName)){
--- a/classes/ExternalModules.php
+++ b/classes/ExternalModules.php
@@ -1533,9 +1533,9 @@ class ExternalModules
 		return $enabledVersions;
 	}
 
-	private static function shouldExcludeModule($prefix, $version)
+	private static function shouldExcludeModule($prefix, $version = null)
 	{
-		if (strpos($_SERVER['REQUEST_URI'], '/manager/ajax/enable-module.php') !== false && $prefix == $_POST['prefix'] && $_POST['version'] != $version) {
+		if ($version && strpos($_SERVER['REQUEST_URI'], '/manager/ajax/enable-module.php') !== false && $prefix == $_POST['prefix'] && $_POST['version'] != $version) {
 			// We are in the process of switching an already enabled module from one version to another.
 			// Do NOT include the currently enabled version of the module to avoid a class name conflict
 			// for the ComposerAutoloaderInit class (if it hasn't changed between module versions).
@@ -1577,10 +1577,11 @@ class ExternalModules
 				$key = $row['key'];
 				$value = $row['value'];
 
+				if (self::shouldExcludeModule($prefix, $key == self::KEY_VERSION ? $value : null)) {
+					continue;
+				}
+
 				if($key == self::KEY_VERSION){
-					if (self::shouldExcludeModule($prefix, $value)) {
-						continue;
-					}
 					$systemwideEnabledVersions[$prefix] = $value;
 				}
 				else if($key == self::KEY_ENABLED){
--- a/classes/ExternalModules.php
+++ b/classes/ExternalModules.php
@@ -399,6 +399,8 @@ class ExternalModules
 	# disables a module system-wide
 	static function disable($moduleDirectoryPrefix)
 	{
+		$version = self::getModuleVersionByPrefix($moduleDirectoryPrefix);
+		self::callHook('redcap_module_system_disable', array($moduleDirectoryPrefix, $version));
 		self::removeSystemSetting($moduleDirectoryPrefix, self::KEY_VERSION);
 
 		// Disable any cron jobs in the crons table
@@ -417,12 +419,25 @@ class ExternalModules
 		self::isCompatibleWithREDCapPHP($moduleDirectoryPrefix, $version);
 
 		if (!isset($project_id)) {
+			$old_version = self::getModuleVersionByPrefix($moduleDirectoryPrefix);
+
 			self::initializeSettingDefaults($instance);
 			self::setSystemSetting($moduleDirectoryPrefix, self::KEY_VERSION, $version);
+
+			self::cacheAllEnableData();
+			if ($old_version) {
+				self::callHook('redcap_module_system_change_version', array($moduleDirectoryPrefix, $version, $old_version));
+			}
+			else {
+				self::callHook('redcap_module_system_enable', array($moduleDirectoryPrefix, $version));
+			}
+
 			self::initializeCronJobs($instance, $moduleDirectoryPrefix);
 		} else {
 			self::initializeSettingDefaults($instance, $project_id);
 			self::setProjectSetting($moduleDirectoryPrefix, $project_id, self::KEY_ENABLED, true);
+			self::cacheAllEnableData();
+			self::callHook('redcap_module_project_enable', array($moduleDirectoryPrefix, $version, $project_id));
 		}
 	}
 
@@ -809,6 +824,11 @@ class ExternalModules
 							'$value'
 						)";
 			} else {
+				if ($key == ExternalModules::KEY_ENABLED && $value == "false" && $pidString != "NULL") {
+					$version = self::getModuleVersionByPrefix($moduleDirectoryPrefix);
+					self::callHook('redcap_module_project_disable', array($moduleDirectoryPrefix, $version, $projectId));
+				}
+
 				$event = "UPDATE";
 				$sql = "UPDATE redcap_external_module_settings
 						SET value = '$value',
@@ -1202,9 +1222,9 @@ class ExternalModules
 			$pid = null;
 			if(!empty($arguments)){
 				$firstArg = $arguments[0];
-				if((int)$firstArg == $firstArg){
+				if (is_numeric($firstArg)) {
 					// As of REDCap 6.16.8, the above checks allow us to safely assume the first arg is the pid for all hooks.
-					$pid = $arguments[0];
+					$pid = $firstArg;
 				}
 			}
 
@@ -1455,7 +1475,7 @@ class ExternalModules
 
 	private static function shouldExcludeModule($prefix, $version)
 	{
-		if(strpos($_SERVER['REQUEST_URI'], '/manager/ajax/enable-module.php') !== false && $prefix == $_POST['prefix']){
+		if (strpos($_SERVER['REQUEST_URI'], '/manager/ajax/enable-module.php') !== false && $prefix == $_POST['prefix'] && $_POST['version'] != $version) {
 			// We are in the process of switching an already enabled module from one version to another.
 			// Do NOT include the currently enabled version of the module to avoid a class name conflict
 			// for the ComposerAutoloaderInit class (if it hasn't changed between module versions).
@@ -1497,11 +1517,10 @@ class ExternalModules
 				$key = $row['key'];
 				$value = $row['value'];
 
-				if(self::shouldExcludeModule($prefix, self::KEY_VERSION)){
-					continue;
-				}
-
 				if($key == self::KEY_VERSION){
+					if (self::shouldExcludeModule($prefix, $value)) {
+						continue;
+					}
 					$systemwideEnabledVersions[$prefix] = $value;
 				}
 				else if($key == self::KEY_ENABLED){
--- a/docs/official-documentation.md
+++ b/docs/official-documentation.md
@@ -82,8 +82,8 @@ Below is a *mostly* comprehensive list of all items that can be added to the  **
 	* A **name** to be displayed on the site
 	* An **icon** in REDCap's image repository
 	* A **url** either in the local directory or external to REDCap.
-* **system-settings** specify settings configurable at the system-wide level (this Control Center). Individual projects can override these settings. 
-* **project-settings** specify settings configurable at the project level, different for each project.  
+* **system-settings** specify settings configurable at the system-wide level (this Control Center). Individual projects can override these settings.  Settings do NOT have to be defined in config.json to be used programatically.  
+* **project-settings** specify settings configurable at the project level, different for each project.  Settings do NOT have to be defined in config.json to be used programatically.  
 * A setting consists of:
 	* A **key** that is the unique identifier for the item. Dashes (-'s) are preferred to underscores (_'s).
 	* A **name** that is the plain-text label for the identifier. You have to tell your users what they are filling out.
--- a/docs/official-documentation.md
+++ b/docs/official-documentation.md
@@ -143,9 +143,48 @@ class HideHomePageEmails extends AbstractExternalModule
 	// Put your code here to get executed by the hook
     }
 }
-````
+```
 
 Remember that each hook function has different method parameters that get passed to it (e.g., $project_id), so be sure to include the correct parameters as seen in the hook documentation for the particular hook function you are defining in your module class.
+
+#### Extra hooks provided by External Modules
+There are a few extra hooks dedicated for modules use:
+
+- `redcap_module_system_enable($prefix, $version)`: triggered when a module gets enabled on Control Center
+- `redcap_module_system_disable($prefix, $version)`: triggered when a module gets disabled on Control Center
+- `redcap_module_system_change_version($prefix, $version, $old_version)`: triggered when a module version is changed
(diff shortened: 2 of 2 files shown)


tbembersimeao:
@mmcev106 do you have plans to to review this PR soon? I totally understand that your team might be busy with priority issues and I don't want to rush you, I just wanted to know an estimate since my team needs this feature to develop an external module.

taylorr4:
@tbembersimeao, given the nature of your PR, i think this one may take @mmcev106 and the Data Core Team a bit more time to consider. What you're proposing is not just a simple code change or modification of functionality, but you're introducing completely new concepts that didn't exist before (i.e., External Modules having their own hooks). While REDCap has hooks, i don't believe i've heard of the idea of EM having its own version of hooks. It's interesting. So while i'm sure these additions will be useful to you and your team, lots of things have to be weighed (like with REDCap itself) to consider whether this all makes sense with regard to the current development path of EM, and whether or not it will be useful to most module developers. 

So in summary, i ask for some patience on this one as i'm sure some things will need to be discussed on the development side (most likely after the holidays) for @mmcev106 and other developers on that team to discern whether this kind of functionality makes sense in the whole grand scheme of things for EM. Thanks!

tbembersimeao:
@taylorr4 thank you for your quick response. Our time frame is flexible, so np at all.

I understand that this not a simple request and it brings questions to be considered, but I just want to advocate one more time that it is important - although not urgent - to provide to developers, either by the proposed approach or not, a way to act on module management events such as enable and disable.

Let's take the use case of a module that needs a new db table to store its own stuff, which is a widespread case in modular web systems like Wordpress and Drupal. Without this feature, the only way to make this happen is to instruct users (sometimes non-technical ones) to run a SQL script manually before enabling the module - which makes the module less modular, less user friendly and more error prone.

FYI my use case is an API client module that needs to log API requests into a new db table.

Again, I do not consider that a top priority for the moment. But as the community grows and new external modules pop up, I believe this can be really useful.

mmcev106:
I totally agree on the importance of this feature.  I will at least give you a status report within the next couple of weeks.

123andy:
Hey @tbembersimeao - re api logging - I have a custom handler that you might find useful:

We use a REDCap project with rules that allow/reject API access based on network, project, or user.  This allows us to open up our VPN networks to everyone but limit external access based on approved projects as a means to mitigate risk a little bit...  You could argue this doesn't do much, but FWIW...

Here is my custom /api/index.php
```
<?php
/*****************************************************************************************
 **  REDCap is only available through ACADMEMIC USER LICENSE with Vanderbilt University
 ******************************************************************************************/

// Set error reporting and prevent caching
error_reporting(E_ALL); //0);

// Prevent caching
header("Expires: 0");
header("cache-control: no-store, no-cache, must-revalidate");
header("Pragma: no-cache");

//Connect to db
include dirname(dirname(__FILE__)) . "/database.php";

$db = new mysqli($hostname, $username, $password, $db);
if ($db->connect_errno) {
    exit("The hostname ($hostname) / username ($username) / password (XXXXXX) combination could not connect to the MySQL server. 
		Please check their values.");
}

/*
 * Taking access control functionality out of apache config file and including it here.
 * For now, if there is no token passed request goes through.
 * If a token is passed, validate request using whitelist rules defined in redcap
 * All successfull and failed requests are logged in the redcap_api_custom_log table
*/

require_once('whitelist.php');
$wl = new whitelist($db);

// IF PASS - EXECUTE REQUEST
if ($wl->evaluate()) {
    
    $file = dirname(dirname(__FILE__)) . "/redcap_v" . $wl->redcap_version . "/API/index.php";

    if (file_exists($file)) {
        $wl->log();
        include $file;
    } else {
        $wl->result = "ERROR";
        $wl->comment = "No REDCap directories found at $file. Please install the REDCap software and try again.";
        $wl->log();
        exit("No REDCap directories found. Please install the REDCap software and try again.");
    }

    // LOG THE COMPLETION OF THE API REQUEST (FOR ACCURATE DURATION)
    // Currently there is a problem that many of the API methods have an 'exit' in them and do not make it back from the include
    // command...  I will try to get these changed so they do return as it allows for much better logging
    $wl->logDuration();
}
```

Here is the whitelist class:
```
<?php

/**
* A php-based whitelist that permits access to the API based on IP and user/project rules
* 
* See https://stanfordmedicine.box.com/s/hjz2ohr4gc47xdzcv6hcn6hp3zpscj0a for the redcap data dictionary
* 
* SYNTAX TO CREATE LOG TABLE

CREATE TABLE `redcap_log_api_custom` (
  `log_id` int(10) NOT NULL AUTO_INCREMENT,
  `ip_address` varchar(50) COLLATE utf8_unicode_ci DEFAULT NULL,
  `username` varchar(255) COLLATE utf8_unicode_ci DEFAULT NULL,
  `project_id` int(10) DEFAULT NULL,
  `ts` timestamp NULL DEFAULT CURRENT_TIMESTAMP,
  `duration` float DEFAULT NULL,
  `result` enum('PASS','REJECT','ERROR') CHARACTER SET utf8 DEFAULT NULL,
  `rule_id` int(10) DEFAULT NULL,
  `comment` text CHARACTER SET utf8,
  PRIMARY KEY (`log_id`)
) ENGINE=InnoDB AUTO_INCREMENT=1 DEFAULT CHARSET=utf8 COLLATE=utf8_unicode_ci;


// Once you have created the REDCap that stores the Whitelist, you can insert it into the config table:

INSERT INTO redcap_config (field_name, value) VALUES ('redcap_log_api_whitelist_pid', XXX);


*/
class whitelist
{
    private $rule_project_id;   // This is the redcap project that contains the rule data dictionary
    private $rules = array();   // Rules pulled from the REDCap Whitelist Project
    public $redcap_version;     // This is the redcap version number

    private $ts_start;          // This is the starting microtime
    private $db;                // This carries the mysqli db

    private $token;             // The API Token
    private $username;          // The TOKEN username
    private $project_id;        // The TOKEN project_id

    private $ip;                // This is the $_REQUEST ip address
    private $rule_id;           // The matching rule_id for logging

    public $log_id;
    public $result = "ERROR";   // PASS, REJECT, or ERROR
    public $comment;            // An optional annotation to be added to the log

    public $rejection_message = <<<EOT
ERROR: Stanford REDCap API Access Restricted

API Access (including the REDCap Mobile App) from untrusted networks requires:

1) an active VPN connection (see https://vpn.stanford.edu for VPN instructions)
    OR
2) an approved project/user-based API Exemption.

To connect now, confiure a VPN connection on your device or if there is a reason this is not possible,
contact us for more information at redcap-help@lists.stanford.edu or review https://medwiki.stanford.edu/x/hAtrAg
EOT;

    public function __construct(mysqli $db)
    {
        $this->ts_start = microtime(true);
        $this->ip = trim($_SERVER['REMOTE_ADDR']);
        $this->db = $db;
        
        $this->getRuleProjectId();
        $this->getRedcapVersion();
        $this->loadRules();
        $this->loadToken();
    }


    public function getRedcapVersion() {
        $sql = "SELECT value FROM redcap_config WHERE field_name = 'redcap_version'";
        if (!$result = $this->db->query($sql)) { die("Error running query at " . __LINE__ . " of " . __FILE__); }
        if ($result->num_rows == 0) { die("ERROR: redcap_version is not configured"); }
        $row = $result->fetch_assoc();
        $this->redcap_version = $row['value'];
    }

    
    public function getRuleProjectId() {
        $sql = "SELECT value FROM redcap_config WHERE field_name = 'redcap_log_api_whitelist_pid'";
        if (!$result = $this->db->query($sql)) { die("Error running query at " . __LINE__ . " of " . __FILE__); }
        if ($result->num_rows == 0) { die("ERROR: redcap_log_api_whitelist_pid is not configured"); }
        $row = $result->fetch_assoc();
        $this->rule_project_id = $row['value'];
    }
    

    public function evaluate() {
        // Step 1: Is the token missing
        if (empty($this->token)) {
            $this->comment  = "[No Token]: " . json_encode($_REQUEST);
            $this->result   = "PASS";
            return true;
        }

        // Load the project and username from the token
        $this->loadProjectUsername();
          
        // Step 2: Is the IP Whitelisted
        if (self::validIpAddress()) {
            $this->result    = "PASS";
            return true;
        }

        // Step 3: Is the username/project_id whitelisted
        if (self::validProjectUsername()) {
            $this->result    = "PASS";
            return true;
        }

        // Step 4: Failed all rules - reject
        $this->result        = "REJECT";
        self::log();

        header('HTTP/1.0 401 Unauthorized');
        exit($this->rejection_message . PHP_EOL . "If you are connected to the VPN and still receiving this message, please " .
            "contact us and include your current IP (" . $this->ip . ").");
    }


    public function validIpAddress() {
        foreach ($this->rules as $rule) {
            if (!empty($rule['ip_address'])) {
                if (self::ipCIDRCheck($rule['ip_address'])) {
                    $this->rule_id = $rule['rule_id'];
                    return true;
                }
            }
        }
        return false;
    }


    public function validProjectUsername() {
        // Skip if we don't have a valid username and project_id
        if (empty($this->username) OR empty($this->project_id)) return false;
        
        // Go through each rule and check for a match
        foreach ($this->rules as $rule) {
            if (empty($rule['username']) AND empty($rule['project_id'])) continue;
            
            if ($rule['username'] === $this->username AND $rule['project_id'] == $this->project_id) {
                // Allow based on username and project_id match
                $this->rule_id = $rule['rule_id'];
                return true;
            }
            
            if ($rule['project_id'] == $this->project_id AND empty($rule['username'])) {
                // Allow based on project_id
                $this->rule_id = $rule['rule_id'];
                return true;
            }
            
            if ($rule['username'] === $this->username AND empty($rule['project_id'])) {
                // Allow based on username role
                $this->rule_id = $rule['rule_id'];
                return true;
            }
        }
        return false;
    }


    public function loadRules() {
        $sql = "
            SELECT * 
            FROM (
                SELECT
                    a.record as rule_id,
                    max(if(a.field_name = 'username', a.value, NULL)) as username,
                    max(if(a.field_name = 'project_id', a.value, NULL)) as project_id,
                    max(if(a.field_name = 'ip_address', a.value, NULL)) as ip_address,
                    max(if(a.field_name = 'inactive', a.value, 0)) as inactive
                FROM
                    redcap_data a
                WHERE
                    a.project_id = " . intval($this->rule_project_id) . " 
                    AND a.record <> ''
                GROUP BY
                    a.record
            ) b
            WHERE
            b.inactive != 1;
        ";
        if (!$result = $this->db->query($sql)) { die("Error running query at " . __LINE__ . " of " . __FILE__); }
        
        while ($row = $result->fetch_assoc()) {
            $this->rules[] = $row;
        }
    }


    public function loadToken() {
        $this->token = isset($_REQUEST['token']) ? trim($_REQUEST['token']) : "";
    }


    public function loadProjectUsername() {
        $sql = "
            SELECT username, project_id 
            FROM redcap_user_rights
            WHERE api_token = ?";
        $stmt = $this->db->prepare($sql);
        $token = trim($this->token);
        $stmt->bind_param('s', $token);
        $stmt->execute();
        $stmt->store_result();
        if($stmt->num_rows > 0) {
            $stmt->bind_result($username, $project_id);
            while ($row = $stmt->fetch()) {
                $this->username = $username;
                $this->project_id = $project_id;
            }           
        } else {
            $this->comment  = "[Invalid Token]: " . json_encode($_REQUEST);
        }
        $stmt->free_result();
        $stmt->close();
    }


    // Checks if the IP is valid given an IP or CIDR range
    // e.g. 192.168.123.1 = 192.168.123.1/30
    public function ipCIDRCheck ($CIDR) {
        $ip = $this->ip;
        if(strpos($CIDR, "/") === false) $CIDR .= "/32";
        list ($net, $mask) = explode("/", $CIDR);
        $ip_net  = ip2long($net);
        $ip_mask = ~((1 << (32 - $mask)) - 1);
        $ip_ip = ip2long($ip);
        $ip_ip_net = $ip_ip & $ip_mask;
        return ($ip_ip_net == $ip_net);
    }

    // Since not all API commands are returning to the calling script, we write the log and update the duration later if it returns
    public function logDuration() {
        $duration = microtime(true) - $this->ts_start;
        $sql = "UPDATE redcap_log_api_custom SET duration = ? WHERE log_id = ?";
        $stmt = $this->db->prepare($sql);
        $stmt->bind_param('di', $duration, $this->log_id);
        $stmt->execute();
        $stmt->close();
    }


    public function log()
    {
        //$comments = empty($this->comment) ? implode(",",array_keys($_REQUEST)) : $this->comment;
        // For now, let's do thorough logging, but we can back off after a while if need be.
        $comments = empty($this->comment) ? json_encode($_REQUEST) : $this->comment;
        $sql = "INSERT INTO redcap_log_api_custom SET 
            ip_address = ?,
            username = ?,
            project_id = ?,
            result = ?,
            rule_id = ?,
            comment = ?";
        $stmt = $this->db->prepare($sql);
        $stmt->bind_param("ssisis",
            $this->ip,
            $this->username,
            $this->project_id,
            $this->result,
            $this->rule_id,
            $comments
        );
        $stmt->execute();
        
        // Record the log ID so we can update the duration later (if it returns)
        $this->log_id = $this->db->insert_id;
        $stmt->close();
    }
    
    public function debug() {
        print "<pre>" . print_r($this,true)."</pre>";
    }
}
```

tbembersimeao:
Thank you @mmcev106!

And thank you @123andy for your code. My API integration module is quite done already, but anyways I really appreciate it - it can be useful in the future, since we are in the same scope of security concerns.

classes/ExternalModules.php
```diff
@@ -1455,7 +1475,7 @@ private static function getEnabledModuleVersionsForProject($pid)
 
 	private static function shouldExcludeModule($prefix, $version)
 	{
-		if(strpos($_SERVER['REQUEST_URI'], '/manager/ajax/enable-module.php') !== false && $prefix == $_POST['prefix']){
+		if (strpos($_SERVER['REQUEST_URI'], '/manager/ajax/enable-module.php') !== false && $prefix == $_POST['prefix'] && $_POST['version'] != $version) {
```
mmcev106: Do you happen to remember why this change was necessary?

classes/ExternalModules.php
```diff
@@ -1497,11 +1517,10 @@ private static function cacheAllEnableData()
 				$key = $row['key'];
 				$value = $row['value'];
 
-				if(self::shouldExcludeModule($prefix, self::KEY_VERSION)){
-					continue;
-				}
-
 				if($key == self::KEY_VERSION){
+					if (self::shouldExcludeModule($prefix, $value)) {
```
mmcev106: ...and this.  It seems like this would populate inconsistent enable data, but I suspect I may not have fully grasped the reasoning here yet.

docs/official-documentation.md
```diff
@@ -143,10 +143,51 @@ class HideHomePageEmails extends AbstractExternalModule
...
+- `redcap_module_project_enable($prefix, $version, $project_id)`: triggered when a module gets enabled on a specific project
+- `redcap_module_project_disable($prefix, $version, $project_id)`: triggered when a module gets disabled on a specific project
+
+Examples:
+
+``` php
+<?php
+
+function redcap_module_system_enable($prefix, $version) {
+    if ($this->PREFIX != $prefix) {
```
mmcev106: I think I might be missing something on the goal of checking the prefix here.  Shouldn't $this->PREFIX always match $prefix?  Actually, do we need to pass the prefix to these new hooks at all?

mmcev106:
I finally got around to reviewing this PR, and it looks fantastic!  Thanks very much for submitting it!  I added a couple of comments/questions to specific lines, but I don't expect them to lead to any significant changes (if any).

classes/ExternalModules.php
tbembersimeao: I am relaxing the exclude criteria because I need the module to be loaded at this initial moment - otherwise the enable (or module change) event will be missed, and the hooks won't be called.

This extra condition makes sure that only the old version is excluded, so we are still avoiding possible conflicts (see comments inside the if clause).

classes/ExternalModules.php
tbembersimeao: `shouldExcludeModule()` expects a version _value_. Previously, the version _key_ was being passed. I fixed that because I needed this parameter to relax a condition (see my answer to the first comment) in order to make sure the enabled module is able to react to its own birth.

docs/official-documentation.md
tbembersimeao: Maybe this is a bad design, but all modules are able to react to this event, not only the enabled one. I remember I did that for convenience, in order to make use of `callHook()` functionality without changing it.

If this approach is not good enough, we could change `callHook()` or `startHook()` in order to prevent other modules to execute these hooks when the event is none of their business.

tbembersimeao: I am working on a fix for that, I'll update the PR soon.

classes/ExternalModules.php
mmcev106: I think my main concern here is that moving the exclude call inside the version check means that $projectEnableOverrides and $projectEnabledDefaults will now get populated for excluded modules (where they didn't previously).  Off the top of my head, I don't know if this will result in any actual problems.  It just seems like it makes the cached data logically inconsistent, which seems like something we should avoid if possible.

docs/official-documentation.md
mmcev106: Thanks!  I really like the idea of moving this check to `callHook()` or `startHook()` to abstract this away from module authors.

classes/ExternalModules.php
tbembersimeao: You are right. I've just committed a fix for that.

docs/official-documentation.md
tbembersimeao: Done.

tbembersimeao:
@mmcev106 thanks for the feedback! I've made a few fixes based on your comments.

classes/ExternalModules.php
mmcev106: As is, I believe this will cause the enable request to crash when upgrading modules with composer dependencies if the ComposerAutoloaderInit class hasn't changed between module versions.  I suspect we may need a different solution to get around this problem (perhaps a separate ajax request).  It may take me a while to get around to it, but I'd be happy to find and implement a fix for this.

mmcev106: On second thought, it doesn't seem like this class loading issue would be limited to that composer class...  I might have to do more testing around this case.

tbembersimeao: How can I reproduce this error? I tried to shift versions of a module containing (and including) the same `vendor/autoload.php` file and I got no error, even for the test in which I return `false` all the times on `shouldExcludeModule()`. Am I missing something?

tbembersimeao:
@mmcev106, I asked you a question in one of our outdated code diff threads that might has been missed. Could you please give me a direction on this? Thanks!

mmcev106:
Was it on ComposerAutoloaderInit thread?  If so, I don't have an answer to that one yet.  It's still on my TODO list, but I'm going to have to find some time to investigate it first before I'll have more info.

tbembersimeao:
@mmcev106 yes, I was just making sure that it wasn't lost among the other notifications/threads/issues/PRs. Thanks!

classes/ExternalModules.php
mmcev106: Please disregard my comments here, I believe this change works just fine.  Module loading has changed since that comment was added.  The ComposerAutoloaderInit scenario is no longer an issue.  I've made a note to update that comment after this PR is merged.

mmcev106:
Thanks for the hard work!  This is an awesome addition.

PR #76, opened 2017-12-21
Discoverable setting bug

taylorr4:
Bug fix: The "discoverable" setting for modules (set in the Control Center for a given module) was not quite working correctly, in which a regular user could navigate to the Project Module Manager page in their project and view *all* modules that have been enabled in the system, when instead they should only be able to view the modules that have been made discoverable.

### DIFF
--- a/classes/ExternalModules.php
+++ b/classes/ExternalModules.php
@@ -2463,12 +2463,15 @@ class ExternalModules
 	public static function getDiscoverableModules()
 	{
 		$modules = array();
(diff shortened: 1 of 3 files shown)


taylorr4:
@mmcev106, would you be able to review this quickly and see if we can expedite getting it into master? Our team at Vanderbilt was ready today to start making modules discoverable to end users, but we quickly realized this bug.

mmcev106:
This commit has been merged into testing and cherry-picked into production and master as commit aee9e15.

PR #79, opened 2018-01-09
UI State

lsgs:
New methods added to AbstractExternalModule that (officially) enable modules to interact with the UI state config via UIState class methods.


taylorr4:
Awesome, thanks, @lsgs! If this looks good to @mmcev106, then i'll let him use his discretion to merge this in.

mmcev106:
I have some concerns about supporting these new methods.  I think there would be unexpected behavior if more that one module was attempting to make UIState calls at the same time (since module execution order is undefined by default).  The order could be controlled by coordinated calls to delayModuleExecution(), but that doesn't seem realistically enforceable.  I'm a bit hesitant to include any API calls in AbstractExternalModule where consistent behavior can't be guaranteed when other modules are enabled.  What does everyone else think?

biggeeves:
I agree with mmcev106.  Unexpected behavior is to be avoided.

taylorr4:
Those are good points. So in the meantime, i think that is fine to leave it as is. @lsgs, that means you will just need to utilize the UIState class directly in your module, which i think should be fine going forward.

lsgs:
Hello all,

That’s all good. I’ve built some methods into my module that are working as I need.

A bit of a query though. I see how you could get unpredictable behaviour with two modules accessing the same UI state object/key due to the sequence of access being undefined. Is this still a concern in this implementation where the methods restrict a module to accessing only UI state settings that belong to that module (because the “object” is the module namespace/name)?

I’m suspecting that an interaction between two modules is going to throw up one or two “interesting features” before too long!

Regards,
Luke

mmcev106:
I apologize, I may have misinterpreted the way the UIState calls work (I haven't used them before).  @taylorr4, whenever you have a minute, can you verify whether the proposed methods would or would not be safe when used by multiple modules within the same hook on the same project?

taylorr4:
@mmcev106, thinking more about this, there's probably not too much to worry about there. The UIState is basically like a specifically-structured type of session array (sort of), except it's kept permanently stored on the user's account. It's typically used for remembering when certain things in the UI are collapsed, hidden, etc. to keep that state persisting indefinitely in the UI for the user.

So after looking at it now, i think that even running multiple hooks/modules in various orders with this, i really can't see it causing any problems. So i think it's actually okay if you want to merge it. To my knowledge, no one else has started to use UIState in plugins/modules yet, but i definitely see it as being useful in the future.

123andy:
I’d love to have a quick example or like to good article.  I have a number of ‘state’ like issues that I currently use custom cookies for - is UI state better?  Will UI state work for unauthenticated survey sessions?

-Andy

(sent from mobile)

On Jan 16, 2018, at 7:54 AM, Rob Taylor <notifications@github.com<mailto:notifications@github.com>> wrote:


@mmcev106<https://github.com/mmcev106>, thinking more about this, there's probably not too much to worry about there. The UIState is basically like a specifically-structured type of session array (sort of), except it's kept permanently stored on the user's account. It's typically used for remembering when certain things in the UI are collapsed, hidden, etc. to keep that state persisting indefinitely in the UI for the user.

So after looking at it now, i think that even running multiple hooks/modules in various orders with this, i really can't see it causing any problems. So i think it's actually okay if you want to merge it. To my knowledge, no one else has started to use UIState in plugins/modules yet, but i definitely see it as being useful in the future.

taylorr4:
@123andy, i think @lsgs could probably show some examples where he's using it in a module. But in general, UI state is a user-account-level attribute (stored as JSON in redcap_user_information), so it won't work for survey sessions because the survey user is not a real user. I would say that UI state is better than cookies because cookies are device-specific, whereas UI state is in the db and therefore is device independent.

lsgs:
@123andy you can take a look at the [DAG Switcher module repo](https://github.com/lsgs/redcap-dag-switcher). It's a pretty minor thing in the context of the module, but I just wanted a way to remember whether a user prefers to see the table with rows as DAGs/columns as users or with rows as users/columns as DAGs.

image

mmcev106:
Thanks very much for the input everyone.  As always, I'm trying to think about changes in terms of long term support, which leads me down the following train of thought (slash rabbit hole):

- To make sure these methods are publicly supported long term, we should add them to the documentation.
- The documentation should say something about them not working in surveys.  On second thought, we should probably implement something that makes them fail gracefully **when** someone tries to use them in a survey.  Should we throw an exception, or fail silently?  Probably throw an exception, since it would most likely get thrown during development, and would make the issue immediately clear.
- Is there anything else we can do to improve clarity?  Maybe the method names should be based on the functionality provided, as opposed to the way they happen to be implemented under the hood.  What if we named the methods **setUserSetting()**, **getUserSetting()**, and **removeUserSetting()** instead?  I think I like that a lot...  Those make much more sense in the list alongside **setSystemSetting()**, **setProjectSetting()**, etc.  That change also makes it more intuitive that the methods will only work in the context of a user (and not in surveys).

What are you thoughts?

taylorr4:
All those points make complete sense to me, @mmcev106.

mmcev106:
Unless anyone has any further thoughts, I will plan to merge this once the described changes are completed.  @lsgs, if you are able and would like to make these changes, please feel free.  If not, I will get around to making them at some point (likely a month or two).

mmcev106:
Thanks for the updated method names.  Do you happen to know how these calls will fail if made from a survey or non-project page?  I think that's my only outstanding concern.  We should also add them to the documentation, but that shouldn't take long.

lsgs:
Thanks for the changes @mmcev106 - all checking out well. To answer your earlier questions:

- On a non-project page the methods work just as they do with a project page - and just as expected. The object is added to an element of the ui-state array under a blank project_id key and gets serialized and written to to redcap_user_information as normal.
image

- On a survey page the ui-state global is updated but the update of redcap_user_information is attempted using a USERID value of "[survey_respondent]", which updates 0 records. This means the ui-state setting persists only within the execution of the current script (via $_GLOBALS['ui-state']) but not across pages. The same goes for NOAUTH pages where USERID is just the string "USERID".

We should include a note to the documentation page saying that the "UserSetting" methods work only in a logged-in context, and not for surveys or NOAUTH pages.

mmcev106:
I went ahead and merged this and added blurb about surveys and NOAUTH pages to the docs:

https://github.com/vanderbilt/redcap-external-modules/commit/d1cc8f48e329bb2d9ced7214a89c3fb244d978e5

Thanks for the up front work @lsgs!  Feel free to make any additional tweaks if you like, and I'll merge them ASAP.

PR #82, opened 2018-02-08
Fix issue with isManagerUrl()

taylorr4:
Fix issue with isManagerUrl() using a server variable that is dependent upon Apache and also on specific versions of Apache


mmcev106:
Awesome!  Thank you @taylorr4 !

PR #83, opened 2018-02-16
Introducing externalModules:configModalReady jQuery event

tbembersimeao:


docs/official-documentation.md
```diff
@@ -341,6 +341,25 @@ If your module will be using JavaScript, it is *highly recommended* that your Ja
...
+
+Besides the custom fields you can implement, you may also listen to a jQuery event - `externalModules:configModalReady` - that is triggered when the configuration modal is fully rendered. The module prefix is provided as parameter. See the example below.
+
+``` javascript
+$(document).ready(function() {
+    var $modal = $('#external-modules-configure-modal');
+
+    $modal.on('externalModules:configModalReady', function(event, modulePrefix) {
+        // Checking whether the current modal is my module's modal.
+        if (modulePrefix !== 'myModulePrefix') {
```
mmcev106: One of our primary goals so far has been to minimize any risks of a bug in one module affecting the functionality of another.  I think there might be a way to implement a JS framework for modules to make this check automatic (removing any risk of it being omitted or mistyped).  We already have a global ExternalModules javascript object which could be used as a starting point.  I have a few ideas which I should be able to explore within the next couple of months.  Let me know if you'd prefer to explore anything before I get around to it.

tbembersimeao: @mmcev106 we could create an additional setting on `config.json`, analogous to `functionName` & `source` for individual fields. Thoughts? Give me a go ahead if you are ok with that.

Example:
``` json
"configModalReadyCallback": {
    "functionName": "myObject.myfunctionName",
    "source": "js/myJsFile.js"
}
```

mmcev106: There's a good chance we might be adding support for a javascript framework including a client side module object in the next couple of weeks (for an unrelated project).  Once that's in place, it should be much easier to make client side features customizable (while minimizing risk of conflict between modules).  It might be prudent to wait on this until that frameworks is completed.

123andy: I like the idea of being able to hijack the config button...  This is one good method that could prove useful.  A second is being able to easily make ajax calls back to the EM (for example to populate dropdowns based on some custom logic).  A third would be to override the entire 'config' page with a custom php page instead of the modal...

mmcev106: A client side module object is still in the works, in case that's helpful for future reference.

Issue #26, opened 2017-08-31
adding javascript / functions to config


123andy:
So I just did a little looking into the use of "source" and "function" as params in config.json.

1) It seems that the 'function' param is only valid on a type=custom.  I can think of cases where you might want to augment an existing field of set type with some custom js so wondering if we should allow functions for any field and not only custom fields.  Downside?

2) Should there be a way to have a 'global' function that is called when the module config is rendered (e.g. after the contents are updated?)  Since the 'click' event for configure on a given module is an anonymous function there is no way to easily hijack it and insert your own processing code.

3) I see that there is some use of the built-in REDCap validation syntax for max and min - can the documentation/examples include this?

123andy:
4. Any thoughts about 'branching' logic for options in the module config?  I would like to offer a checkbox to 'enable' a more advance set of options and would like to hide them if not needed... Could be done with some javascript...

mmcev106:
I have not spent much time considering features like those, but I could see all of them being very useful.  We have not encountered any client needs for them unfortunately, so I'm not sure if/when they would become a priority for my team.

tbembersimeao:
@mmcev106 Our team has needed item 2 a few times already. I just created a PR for that.

tbembersimeao:
@mmcev106 [PR](https://github.com/vanderbilt/redcap-external-modules/pull/84) for item 1 has been created as well.

tbembersimeao:
@mmcev106 I've also made a [PR](https://github.com/vanderbilt/redcap-external-modules/pull/85) for item 4, which has also been a internal request in my team.

tbembersimeao:
I'd like to also add one more item to the list.

5. Create an option to add a helper text to config fields, just like the built in fields.
![helper_texts](https://user-images.githubusercontent.com/29384017/36435836-b95d8984-1630-11e8-8a8d-adf6291cedad.png)

mmcev106:
Closing for lack of updates.  New ideas on the topic would likely take a different approach anyway.

PR #84, opened 2018-02-16
Expanding functionName to all config field types

tbembersimeao:
Proposes a solution for item 1 of @123andy's comment [here](https://github.com/vanderbilt/redcap-external-modules/issues/26#issue-254459527).


mmcev106:
Thanks!

PR #94, opened 2018-04-10
fixed issue with passing image files through api url

123andy:
The current 'require_once' for images doesn't work (even though the content-type is set.  The size is not set and the file fails.  I modified it to pass through all files that are not php or md so you get a more accurate response from the webserver.

### DIFF
--- a/index.php
+++ b/index.php
@@ -50,26 +50,45 @@ if(!file_exists($pagePath)){
 	throw new Exception("The specified page does not exist for this module. $pagePath");
 }
 
-if($pageExtension == 'md'){
-	$Parsedown = new \Parsedown();
-	$html = $Parsedown->text(file_get_contents($pagePath));
+switch ($pageExtension) {
+    case "php":
+    case "":
+        // PHP content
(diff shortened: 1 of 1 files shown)


index.php
```diff
@@ -50,26 +50,45 @@
...
+        header("Content-Type: $mime_type");
+        header('Content-Length: ' . filesize($pagePath));
 
-	// This variable is not used here, but is intended for use inside the file required below.
-	$module = ExternalModules::getModuleInstance($prefix, $version);
-	require_once $pagePath;
+        // stream the file
+        $fp = fopen($pagePath, 'rb');
+        fpassthru($fp);
+        exit();
```
mmcev106: Overall this change looks great.  I'm guessing this exit() call might just be a carry over from a copy paste, but is there some reason it is required?

mmcev106:
I would like to go ahead and merge this.  Would you mind making a pull request for the testing branch instead?

123andy:
going to re-do pull request to test.

PR #96, opened 2018-04-12
do not show the external mod panel if there are no visible links

123andy:
with the redcap_module_link_check_display hook, it is possible that the links defined are all being hidden - in which case we shouldn't display the external modules panel at all on the sidebar.


mmcev106:
Thanks much!

Issue #97, opened 2018-04-18
Branching logic doesn't work with checkboxes?


eaneuhaus:
I'm trying to use the new config page branching logic and it works fine as outlined in the example (branching based on a text field value) but seemingly not with a checkbox. Is there an issue with my logic?

```
{
	"key": "example-checkbox",
	"name": "Enable to see hidden config field",
	"type": "checkbox",
	"required": false
},
{
	"key": "hidden-config-field",
	"name": "This field should be visible if the box above is checked",
	"type": "text",
	"required": false,
	"branchingLogic": {
		"field": "example-checkbox",
		"value": true
	}
}
```

mmcev106:
I have not tried this with checkboxes, but I would suspect that the "value" should be the numeric value associated with a particular checkbox (like what would appear in an export).

eaneuhaus:
Actually, I think I figured out what's happening: The checkbox value isn't updated in realtime, so it's not checking the branching logic until after the config is saved and reloaded. If I check the box, save, and open the config popup again, I see the "hidden-config-field".

`"value": "true"` is the correct syntax, but this is definitely still an issue.

mmcev106:
I apologize, my first comment didn't make any sense (just deleted it).  I haven't really worked with configuration branching logic yet aside from basic testing.  @tbembersimeao might have a better idea as to what is going on here.

tbembersimeao:
@eaneuhaus @mmcev106 I'll look into it this week. I'll let you know as soon as I have a position about it.

tbembersimeao:
Sorry about my delay on this. I started to work on this issue, and I'll submit a PR later this week, which will also fix #93.

tbembersimeao:
Created a PR to fix it - #98.

PR #98, opened 2018-05-08
Fixing branching logic for checkboxes, radios and sub-settings

### LINKED ISSUE #93: Add support for 'branchingLogic' on 'sub_settings'
The 'branchingLogic' setting option currently only works on top level settings.  It has no affect on 'sub_settings'.

tbembersimeao:
Fixes #93 and #97.

Review steps:
- [ ] Make sure branching logic works when one of the condition fields is a checkbox
- [ ] Make sure branching logic works when one of the condition fields is a radio
- [ ] Make sure branching logic works when one of the condition fields is a sub-setting
- [ ] Make sure branching logic works when the target field is sub-setting
- [ ] Make sure there are no conflicts when branching logic is applied to parent and child fields at the same time:
  1. when parent is hidden, child is hidden no matter what
  2. when the parent is shown, it does not overlap child visibility rules

### DIFF
--- a/manager/js/globals.js
+++ b/manager/js/globals.js
@@ -132,14 +132,35 @@ ExternalModules.Settings.prototype.processBranchingLogicCondition = function(con
 		return false;
 	}
 
-	var $field = $('#external-modules-configure-modal [name="' + condition.field + '"]');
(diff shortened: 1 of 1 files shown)


mmcev106:
Thanks very much!

PR #99, opened 2018-05-10
Fix for download module issue

taylorr4:
@mmcev106, i think this fix is much better than my previous one this morning, which i really should have tested a LOT more and i think may have been responsible for the issues seen on the server earlier. Please take a lot to see if this makes sense. Essentially, the issue i was having is that i would download a module from the Repo, and then immediately when i would open the disabled modules popup to enable it, it would immediately delete the module directory.


mmcev106:
I'm not sure I really understand what is going on here.  The $MODULES_PATH var gets initialized the first time any hook is called.  That makes me think getModuleDirectoryPath() is getting called in some context that is not triggered by a hook.  I'd love to see a stack trace from a call to getModuleDirectoryPath() where $MODULES_PATH is not set.

PR #108, opened 2018-08-16
Emergency patch that is breaking REDCap installations on PHP 5.4 and 5.5

taylorr4:


taylorr4:
const RESERVED_PARAMETER_NAMES is not compatible with PHP 5.4-5.5. Recommend changing it to public static $RESERVED_PARAMETER_NAMES. I chose "public" since it is referenced in AbstractExternalModuleTest.php.

taylorr4:
Note: I went ahead and pushed these changes into the latest REDCap releases as a patch. Feel free to modify any of my changes going forward if you all decide to rework things (making it a private static var or whatever). My changes were merely a quick fix. Thanks!

mmcev106:
Thanks very much @taylorr4...  I also merged this branch into the production branch so you can use it for the next release without worry.

taylorr4:
Great, thanks!

PR #114, opened 2018-09-12
setDAG() and addAutoNumberedRecord() were not updating the record list cache table

taylorr4:


taylorr4:
@mmcev106, this should fix the issue with setDAG() and addAutoNumberedRecord() not updating the record list cache table in REDCap 8.7.2+. If this looks good, please go ahead and merge into both testing and prod branches. I will include these specific changes in REDCap 8.7.3. Thanks!

taylorr4:
@mmcev106 additionally, after you merge these changes into prod and deploy it to Vanderbilt REDCap prod, i recommend running the query below, which will destroy any possibly messed-up record caches that may have been affected by this.

```delete from `redcap_record_counts` where project_id in (select distinct project_id from redcap_external_module_settings where `key` = 'enabled' and `value` = 'true' and project_id is not null)```

mmcev106:
Thanks very much for this fix!  Just FYI, I had to add backslashes before the "Records" class name to avoid an error (since it's in the global namespace).  These changes have been tested, merged, and deployed to production!  I also ran the `delete` query to clear the cache.

taylorr4:
Great, thanks!

PR #120, opened 2018-09-19
Made use of the "new" api endpoint to exclude the redcap version from…

mmcev106:
… rich text image urls


mcguffk:
The changes seems fine. What reason is there for this? Are these images being used in other plugins or hardcoded into the DD?

mmcev106:
The Inline Descriptive Popups module is a good example where this was problematic.

mcguffk:
Yeah, I couldn't see where to approve it without immediately merging it in. I approve

PR #129, opened 2018-10-10
Option groups and select2 for field-list config

dr01d3r:
- field-list configurations will now be grouped and sorted by instrument.
- all select controls will leverage select2 for searchability
- select control widths have been normalized for consistency and better display when converted to select2
  - selects with short text options would always get truncated (ellipsis) when converted to select2


manager/css/style.css
```diff
@@ -150,7 +146,7 @@
 }
 
 #external-modules-configure-modal select {
-	max-width: 275px;
+	width: 275px;
```
mmcev106: I looks like this change causes the long form names to scroll instead of making the select field a bit wider.  Was this intentional?

manager/js/globals.js
```diff
@@ -705,6 +723,7 @@ ExternalModules.Settings.prototype.initializeRichTextFields = function(){
 	});
 
 	$(".external-modules-autocomplete-dropdown").select2();
+    $("#external-modules-configure-modal select.external-modules-input-element").select2();
```
mmcev106: This calls select2 on all select fields (not just field lists), and it looks like it breaks project dropdowns (which are  initialized a few lines above).  I'm not opposed to calling select2 on all selects, but that might require additional testing and some changes.

manager/css/style.css
dr01d3r: That was not intentional.  Select elements with short text entries have display issues when converted to select2.  Most of their text turns to ellipses and unreadable.  Long form names would have issues regardless, given this only replaced max-width, correct?

manager/js/globals.js
dr01d3r: I was inclined to only change the field-list; however, I couldn't find a clean way to isolate those elements.  At the time, I saw no negative impact on other config options; though, I did not test project-list.

manager/css/style.css
mmcev106: Here are some screenshots of what I'm seeing:

width
min-width

The odd horizontal scrolling behavior is the only thing I'm concerned about.  I'm not sure I have a strong preference on the width itself.

manager/js/globals.js
mmcev106: Even if it appears to work, I'm not sure initializing select2 fields multiple times is supported.  It could cause unexpected behavior that's not immediately apparent, and could introduce new issues when we upgrade select2 down the road.  I think we can find a solution that avoids this and fixes project lists at the same time.

mmcev106:
@dr01d3r, do you think you might pick this back up at some point?  No rush, I was just curious if it made sense to leave open.

mmcev106:
I think I'm going to close this for now.  Feel free to reopen it at any time.

PR #135, opened 2018-11-13
enhanced control center links

123andy:
Allow the 'redcap_module_link_check_display' hook to work on control center links as well project-specific links.   

An example use of this is to visually queue a super user that a module is not configured correctly by changing the icon to an exclamation, etc...


mmcev106:
Merged.  This should move to the production branch on the 26th, and should be included in the next release after that. Thanks Andy!

PR #150, opened 2018-12-03
Internal module integration (i.e., bundled module functionality)

taylorr4:
This allows for...
1. Modules to be placed as-is in the redcap_vX.X.X/Modules/ directory, after which they will get auto-enabled in REDCap. If a bundled module is already enabled but on a different version, it will be auto-disabled, and the bundled module will be auto-enabled.
1. If an admin navigates to the REDCap Repo, any bundled modules will be hidden in the Repo. Also, bundled modules will not have any module update notifications displayed in the Control Center (if a new version happens to be released in the Repo).
1. Any bundled modules will not get displayed on the External Modules page in the Control Center (because they can neither be disabled nor enabled and their version cannot be changed through the user interface - they are an integrated part of the system).
1. If a developer manually pushes a new version of a module into the redcap_vX.X.X/Modules/ directory, in which multiple versions of the same module would exist in that directory, then REDCap will assume the one with the highest version to be the correct one (and if not enabled, will auto-enable it). This will allow developers to upgrade to new module versions in case of an emergency module update.
1. Ideally going forward, any bundled modules will not have any system-level configurations to set, but if they do, the default values for the system-level configuration will be set for the module when the module gets auto-enabled.

@mmcev106, let me know if you have any comments, questions, or concerns about these.

### DIFF
--- a/classes/ExternalModules.php
+++ b/classes/ExternalModules.php
@@ -93,6 +93,8 @@ class ExternalModules
 	private static $deletedModules;
 
 	private static $configs = array();
+	
+	private static $bundledModules;
 
 	# two reserved settings that are there for each project
 	# KEY_VERSION, if present, denotes that the project is enabled system-wide
@@ -247,10 +249,10 @@ class ExternalModules
 	{
 		// Get module directories
 		if (defined("APP_PATH_EXTMOD")) {
-			$modulesDirectories = [dirname(APP_PATH_DOCROOT).DS.'modules'.DS, APP_PATH_EXTMOD.'example_modules'.DS];
+			$modulesDirectories = [APP_PATH_DOCROOT.'Modules'.DS, dirname(APP_PATH_DOCROOT).DS.'modules'.DS, APP_PATH_EXTMOD.'example_modules'.DS];
 		} else {
-			$modulesDirectories = [dirname(APP_PATH_DOCROOT).DS.'modules'.DS, dirname(APP_PATH_DOCROOT).DS.'external_modules'.DS.'example_modules'.DS];
-		}		
+			$modulesDirectories = [APP_PATH_DOCROOT.'Modules'.DS, dirname(APP_PATH_DOCROOT).DS.'modules'.DS, dirname(APP_PATH_DOCROOT).DS.'external_modules'.DS.'example_modules'.DS];
+		}
 		// Add any alternate module directories
 		$modulesDirectoriesAlt = self::getAltModuleDirectories();
 		foreach ($modulesDirectoriesAlt as $thisDir) {
@@ -269,11 +271,38 @@ class ExternalModules
 		foreach ($modulesDirectories as $dir) {
 			foreach (getDirFiles($dir) as $module) {
 			    // Use the module directory as a key to prevent duplicates from alternate module directories.
-				$modules[$module] = true;
+				if (is_dir($dir.$module)) {
+					$modules[$module] = true;
+				}
 			}
 		}
 		// Return directories array
 		return array_keys($modules);
+	}
+
+	// Return boolean regarding if the module is a bundled module
+	public static function isBundledModule($prefix)
+	{
+		// Initialize as array
+		self::initBundledModules();
+		// Is it in the array?
+		return isset(self::$bundledModules[$prefix]);
+	}
+
+	// Initialize the bundled modules array
+	public static function initBundledModules()
+	{
+		if (is_array(self::$bundledModules)) return;
+		// Initialize as array
+		self::$bundledModules = array();	
+		// Get all module directorys in the redcap_vX.X.X/Modules/ directory
+		$dir = APP_PATH_DOCROOT.'Modules'.DS;
+		if (!is_dir($dir)) return;
+		foreach (getDirFiles($dir) as $folder) {
+			if (!is_dir($dir.$folder)) continue;
+			list ($thisPrefix, $thisVersion) = self::getParseModuleDirectoryPrefixAndVersion($folder);
+			self::$bundledModules[$thisPrefix] = $thisVersion;
+		}
 	}
 
 	# initializes the External Module aparatus
@@ -760,8 +789,33 @@ class ExternalModules
 		}
 	}
 
+	static function getSettingDefaultFromConfig($moduleDirectoryPrefix, $key, $systemSetting=true)
+	{
+		$settingsName = $systemSetting ? 'system-settings' : 'project-settings';
+		$config = self::getConfig($moduleDirectoryPrefix);
+		if (!isset($config[$settingsName])) return null;
+		foreach ($config[$settingsName] as $attr) {
+			if ($attr['key'] != $key) continue;
+			return isset($attr['default']) ? $attr['default'] : null;
+		}
+		return null;
+	}
+
 	static function getSystemSetting($moduleDirectoryPrefix, $key)
 	{
+		// For bundled modules, their systems will not be stored in the db table
+		if (self::isBundledModule($moduleDirectoryPrefix)) {
+			// If asking for the version of a module that is bundled, always return the 
+			// bundled version number (in case other versions exist elsewhere)
+			if ($key == self::KEY_VERSION) {
+				return self::$bundledModules[$moduleDirectoryPrefix];
+			} 
+			// Other module attribute from default in config.json
+			else {
+				return self::getSettingDefaultFromConfig($moduleDirectoryPrefix, $key, true);
+			}
+		}
+		// Return setting from db table
 		return self::getSetting($moduleDirectoryPrefix, self::SYSTEM_SETTING_PROJECT_ID, $key);
 	}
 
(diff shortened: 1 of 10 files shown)


classes/ExternalModules.php
```diff
@@ -1721,23 +1791,66 @@ private static function cacheAllEnableData()
...
 			while($row = self::validateSettingsRow(db_fetch_assoc($result))) {
+				// Add db row values to arrays
 				$key = $row['key'];
 				if ($key == self::KEY_VERSION) {
 					$result_versions[] = $row;
+					$prefix_version[$row['directory_prefix']] = $row['value'];
 				} else if($key == self::KEY_ENABLED) {
-					$result_enabled[] = $row;
+					if (!self::isBundledModule($row['directory_prefix'])) { // If bundled, we'll set it as enabled below
+						$result_enabled[] = $row;
```
mmcev106: What is the reason for changing the project enable flag handling?  It seems like these would work without any changes.

classes/ExternalModules.php
```diff
@@ -2207,6 +2321,9 @@ public static function getAdditionalFieldChoices($configRow,$pid) {
 	public static function getEnabledVersion($prefix)
 	{
 		$versionsByPrefix = self::getSystemwideEnabledVersions();
+		if (!isset($versionsByPrefix[$prefix]) && isset(self::$bundledModules[$prefix])) {
+			$versionsByPrefix[$prefix] = self::$bundledModules[$prefix];
```
mmcev106: Is this necessary?  It seems like the correct version would already have been cached in cacheAllEnableData().

manager/templates/enabled-modules.php
```diff
@@ -42,7 +42,7 @@
 			<div class="modal-header">
 				<h4 class="modal-title clearfix">
 					<div class="float-left">Available Modules</div>
-					<div class="float-right" style="margin-left:50px;"><input type="text" id="disabled-modules-search" class="quicksearchsm" placeholder="Search available modules" autofocus></div>
```
mmcev106: Why was "autofocus" removed?  I actually use that feature a lot...

manager/templates/enabled-modules.php
```diff
@@ -168,9 +168,16 @@
...
 		<input type="hidden" name="redcap_version" value="<?=REDCAP_VERSION?>">		
 		<input type="hidden" name="institution" value="<?=htmlspecialchars($GLOBALS['institution'], ENT_QUOTES)?>">
-		<?php foreach (\ExternalModules\ExternalModules::getModulesInModuleDirectories() as $thisModule) { ?>
-			<input type="hidden" name="downloaded_modules[]" value="<?=$thisModule?>">
-		<?php } ?>
+		<?php 
+		foreach (\ExternalModules\ExternalModules::getModulesInModuleDirectories() as $thisModule) {
+			?><input type="hidden" name="downloaded_modules[]" value="<?=$thisModule?>"><?php 
+		}
+		if (method_exists('\ExternalModules\ExternalModules', 'getBundledModulePrefixes')) {
```
mmcev106: What is the scenario where getBundledModulePrefixes() doesn't exist?

classes/ExternalModules.php
mmcev106: I just tested moving a module already enabled on a project to the "Modules" folder and the module then acted as if it was disabled on the project, but did not appear in the available module list.  This might not be working quite right...

classes/ExternalModules.php
```diff
@@ -1721,23 +1791,66 @@ private static function cacheAllEnableData()
...
+					if (!self::isBundledModule($row['directory_prefix'])) { // If bundled, we'll set it as enabled below
+						$result_enabled[] = $row;
+					}
 				} else {
 					throw new Exception("Unexpected key: $key");
 				}
 			}
+			
+			## AUTO-ENABLE BUNDLED MODULES
+			// Skip enabling of bundled module if we're recursively trying to enable it right now
```
mmcev106: Interesting...  That that recursive call an expected use case, or something we should consider addressing?

classes/ExternalModules.php
mmcev106: This set of changes could cause some very odd behavior for anyone currently using the "Enable module on all projects by default" system setting when they install an update that bundles that module.  There's a good chance nobody uses that setting though.  Perhaps we should ask the community, and then just remove that setting entirely to remove the risk going forward.

manager/templates/enabled-modules.php
taylorr4: Sorry, not sure how that happened. Was a mistake. I've re-added it.

classes/ExternalModules.php
taylorr4: I honestly don't know how else to address it. There's so much recursive stuff in the code, it's hard for me to figure out how to insert certain actions. This was the best way i could think of to deal with it that worked.

classes/ExternalModules.php
taylorr4: @mmcev106 Yes, it seems to be necessary, because without it,  a lot of recursive processing inside cacheAllEnableData() would sometimes cause the bundled module to not be found during the process of auto-enabling *some*  modules.

classes/ExternalModules.php
```diff
@@ -764,8 +806,36 @@ static function initializeSettingDefaults($moduleInstance, $pid=null)
...
+		foreach ($config[$settingsName] as $attr) {
+			if ($attr['key'] != $key) continue;
+			return isset($attr['default']) ? $attr['default'] : null;
+		}
+		return null;
+	}
+
 	static function getSystemSetting($moduleDirectoryPrefix, $key)
 	{
+		// For bundled modules, their systems will not be stored in the db table
```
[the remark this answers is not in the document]
taylorr4: You're right. This change was early on in my work, and looking at it now, it isn't needed. I've removed this code now and also made sure that the bundled module gets enabled on all projects by default, which was lacking in previous commits.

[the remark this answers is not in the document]
mmcev106: What do you think about still having to enable bundled modules on each project?  To me, that seems prudent simply for splitting up risks like conflicts with other modules, or bugs that could bring down a whole system instead of a few projects (ex: a performance issue in save_record).  That seems in line with having to enable "internal" modules like repeating instruments.
 It also looks like it might make this set of changes a little simpler and more maintainable.

classes/ExternalModules.php
mmcev106: The odd recursion might be introduced by calling enableAndCatchExceptions() inside cacheAllEnableData().  I will plan to think through this a little more.

classes/ExternalModules.php
mmcev106: That would suggest that THE core method that determines when modules should execute is not working consistently, which I think definitely warrants investigation.  I will comment that line and try to reproduce the issue.

classes/ExternalModules.php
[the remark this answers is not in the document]
taylorr4: I see your point about performance and it being similar to enabling certain interal features now, although i'm not quite sure how bundled modules could conflict with other modules given that we're using namespacing (but maybe you mean some other way). 

But i'm not sure how i feel about having to enable them on a project-by-project basis.  I'd argue that the "internal" modules in REDCap are already enabled in all projects until the user chooses to use them, so i would opt more for leaving bundled modules enabled on all projects by default. Since we ourselves are bundling these into the code, i feel fairly confident that we shouldn't have issues here with performance or other things because we're controlling this (and if we do, we can always modify the individual module to improve these things, right?).

classes/ExternalModules.php
taylorr4: Yeah, i'm pretty sure that's the case.

classes/ExternalModules.php
taylorr4: Okay, as a note, try it with a few different modules. For instance, i don't have any issue with it on most modules, but if i try to bundle Survey Link Lookup  module, it fails to add the version row to the module settings db table. I can't figure out why that module is giving trouble, but other ones (including system-level only modules) don't have this issue.

taylorr4: Apparently, it seems to be an issue if the module has required system settings, in which case initializeSettingDefaults() seems to fail mid-way through. I don't think this is a real issue but just a random edge case and an idiosyncrasy. Since we are ultimately controlling and testing the modules being bundled, i don't see this an a real issue. So i've removed that code i added to getEnabledVersion().

classes/ExternalModules.php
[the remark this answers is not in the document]
mmcev106: I think people are mostly namespacing their modules, but there is still a risk of two people defining the same thing in the global namespace, both in modules themselves as well as legacy hooks.  For example, if any older Vandy projects require PHPMailer explicitly, the Email Alerts module could load that class first, causing REDCap to crash on the explicit require in the old hook.  There are probably other examples...  I'd have to dedicate more time to thinking this through before speculating on the size of that risk.

What are your thoughts?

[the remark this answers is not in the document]
taylorr4: Wow, i honestly had not thought of the the third-party stuff like PHPMailer or anything inside vendor for a given module. What if 2 separate modules were using PHPMailer and both were installed and running on the same page? Wouldn't they conflict? If so, then this is a real issue right now then, right? I wasn't aware of this and honestly not sure how it might be dealt with. Ideas?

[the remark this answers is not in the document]
mmcev106: This is why we have the language in the docs about using composer for third party dependencies.  Using a classloader replaces conflicts with a "first version loaded wins" scenario.  While this still has the potential to cause problems when dependency versions are mismatched, it is often a non-issue in practice.  The WordPress plugin/theme community has the same risks and often uses composer as a solution, which I suspect will work well for us long-term  We've had several talks in the EM community about this, and possibly someday implementing a Drupal-like dependency management system if it does become a larger issue.  There've been multiple community discussions an a couple of GitHub issues that touch on the topic.

From my perspective it is a less common but real risk that has the potential to leave people dead in the water for weeks.  I think it would to prudent to mitigate it by still requiring bundled modules to explicitly be enabled on each project.

mmcev106:
Rob and I since talked in person a lot about these changes.  He's considering dropping this idea (bundled modules) in favor of some other method of promoting top modules.

taylorr4:
Abandoning this idea.

PR #152, opened 2018-12-15
Update all modules semi-automatically

taylorr4:
New button that will update all modules instead of the admin having to update them individually


taylorr4:
Great, thanks! I feel comfortable with it, so I'll go ahead and merge to production then.

PR #158, opened 2019-01-09
Made getSubSettings() work with nested sub settings

mmcev106:
I believe this change should be fully backward compatible as verified by the unit test.  @kcmcg, would you be be able to check me on that?

@tmwil and @tbembersimeao, you may want to review this since I believe you both implemented something similar about a year ago in your respective projects.  I think we just never replaced the getSubSettings() implementation because we weren't sure about the backward compatibility of those implementations.  I ran into this again on my current project and was hoping to solve it once and for all.  Do you think this solution would work for similar cases going forward?

@Pyronaught, don't feel obligated to dig into the details of what's going on here unless you want to.  I mainly thought you'd be curious.

### DIFF
--- a/classes/AbstractExternalModule.php
+++ b/classes/AbstractExternalModule.php
@@ -262,20 +262,30 @@ class AbstractExternalModule
 
 	function getSubSettings($key, $pid = null)
 	{
-		$keys = [];
-		$config = $this->getSettingConfig($key);
-		foreach($config['sub_settings'] as $subSetting){
-			$keys[] = $this->prefixSettingKey($subSetting['key']);
-		}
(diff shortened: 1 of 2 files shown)


mmcev106:
My unit test isn't correct apparently.  It is behavior differently on my projects.  I'm sure more changes will follow if you want to wait to review this.

mmcev106:
I've reverted this for now.  I do think it's backward compatible, but the new functionality doesn't seem to work consistently in all cases (ex: repeatable and/or one vs. multiple sub_setting values).  I think I should write some more comprehensive unit tests before moving forward.

PR #161, opened 2019-01-15
Added mutual exclusion for setSetting()

mmcev106:
**When reviewing these code changes, I would highly recommend checking "Hide whitespace changes" under "Diff settings" at the top.**

The original setSettings() implementation was a prototype and was not expected to live this long without review or refactoring.  It did not offer true mutual exclusion, which has been resulting in more "More than one (2) External Module setting exists" errors and "Unexpected number of affected rows" warnings lately.  I believe this solution is solid, but I have not done anything quite like it before.  Overlooking something could crash REDCap (for us and others).  If possible, would you all be able to review this in depth to hopefully make sure all our bases are covered?  I originally started using transactions, but ran into deadlock issues.  I stumbled across GET_LOCK() while troubleshooting, and it seemed like a more appropriate solution anyway.  Feel free to reach out with any and all questions/concerns.

### DIFF
--- a/classes/ExternalModules.php
+++ b/classes/ExternalModules.php
@@ -879,121 +879,146 @@ class ExternalModules
 	# call set [System,Project] Setting instead of calling this method
 	private static function setSetting($moduleDirectoryPrefix, $projectId, $key, $value, $type = "")
 	{
-		if(self::areSettingPermissionsUserBased($moduleDirectoryPrefix, $key)) {
-			$errorMessageSuffix = "You may want to use the disableUserBasedSettingPermissions() method to disable this check and leave permissions up the the module's code.";
+		$externalModuleId = self::getIdForPrefix($moduleDirectoryPrefix);
+		$lockName = db_real_escape_string("external-module-setting-$externalModuleId-$projectId");
 
-			if ($projectId == self::SYSTEM_SETTING_PROJECT_ID) {
-				if (!defined("CRON") && !self::hasSystemSettingsSavePermission($moduleDirectoryPrefix)) {
-					throw new Exception("You don't have permission to save system settings!  $errorMessageSuffix");
-				}
+		// The natural solution to prevent duplicates would be a unique key.
+		// That unfortunately doesn't work for the settings table since the total length of the appropriate key columns is longer than the maximum unique key length.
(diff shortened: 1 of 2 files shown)


kcmcg:
@taylorr4 Do you know if REDCap will kill duplicate GET_LOCK requests, if 2 people attempt to edit the same setting at the same time? Or will it only kill queries if the request comes from the same IP address?

@mmcev106 From reading up on this, it sounds like during normal operations, GET_LOCK will wait until the previous lock is removed, rather than immediately returning "0" just because 2 users attempt to edit the same setting at the same time. Is that your understanding as well? If so, this sounds like a great solution to me. If not, I worry that this will cause too many error messages and that we'll need to implement a loop to attempt to re-acquire lock if it doesn't get a lock the first try.

mmcev106:
Great questions...

- The unit test runs 1000 iterations on two concurrent processes (2000 iterations total).  Each iteration sets the same setting, then removes it.  It is curious that the query killing hasn't come into play in the unit test...  If it ever did, I believe it would just cause an exception in the setSetting() call, which I think is acceptable.
- Yes, GET_LOCK() just hangs until the lock is free.  This should be a fraction of a second unless the database is completely bogged down.  I chose the five second timeout semi-arbitrarily.  I figure we must have bigger database issues if it's taking longer than five seconds, in which case returning an exception is appropriate.

taylorr4:
I honestly have never used GET_LOCK and know little about the locking functions in MySQL. But from what i've read in the documentation, it sounds good for usage here. Reviewing these changes, the whole approach makes sense to me here, but i would defer to you @mmcev106 with regard to testing of this beforehand. So if you've tested it and feel comfortable with it, then i feel comfortable. Thanks!

mmcev106:
@Pyronaught & @jscherdin, are you able/planning to review this?  Excluding the unit test It's only a couple dozen lines if you check "Hide whitespace changes".  If you're not able that's OK, but I would curious to hear any questions/concerns you might have.

mmcev106:
Via slack @Pyronaught said he reviewed this and didn't see any potential issues.  I am going to go ahead and deploy this to redcaptest with the intent of deploying to prod on 28th.  @jscherdin, please still feel free to chime in if you have any concerns.  Thank you all very much for your time & input.

PR #162, opened 2019-01-18
Take two on making getSubSettings() work with nested sub settings

mmcev106:
I fixed issue I found after the first PR.  I believe this change should be fully backward compatible as verified by the unit test. @kcmcg, would you be be able to check me on that?

@tmwil and @tbembersimeao, you may want to review this since I believe you both implemented something similar about a year ago in your respective projects. I think we just never replaced the getSubSettings() implementation because we weren't sure about the backward compatibility of those implementations. I ran into this again on my current project and was hoping to solve it once and for all. Do you think this solution would work for similar cases going forward?

@Pyronaught, don't feel obligated to dig into the details of what's going on here unless you want to. I mainly thought you'd be curious.

### DIFF
--- a/classes/AbstractExternalModule.php
+++ b/classes/AbstractExternalModule.php
@@ -278,15 +278,47 @@ class AbstractExternalModule
 			if($subSettingConfig['type'] === 'sub_settings'){
 				// Handle nested sub_settings recursively
 				$values = $this->getSubSettings_internal($settingsAsArray, $subSettingConfig);
+				
+				$recursionCheck = function($value){
+					// We already know the value must be an array.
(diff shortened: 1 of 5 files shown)


mmcev106:
@kcmcg, @Pyronaught, @tmwil, are you interested/able to review this?  If not, that's ok.  If so, do you have a time frame in mind?  A response either way would be appreciated as it would help me with planning.

kcmcg:
I just looked through it and did some testing in the new branch. It looks like it's working pretty well. The PDF Modification module is one that has 3 levels deep of nesting with repeating fields at all levels and it was working pretty well on the new branch.

mmcev106:
Awesome, thanks very much for testing @kcmcg.

Issue #89, opened 2018-03-09
Add support for the redcap_custom_verify_username hook and/or hook return values


mmcev106:
See this discussion:

https://community.projectredcap.org/questions/36986/external-modules-version-of-redcap-custom-verify-u.html?childToView=37095

123andy:
I just beat my head on the wall (and total forgot about this thread/issue).  @lsgs 

I couldn't get it working and starting playing around trying to figure out why the 'return' value wasn't getting out of the EM.  I made a few tweaks to the EM code so that it made its way out but found that this still did not solve the problem.  

There was another change made to REDCap (probably some time ago) where the way that this hook is 'handled' - it no longer processes the 'status' and 'message' as is specified in the docs.   Basically, it ignores the return value.

So, here we are:
- I did a little hacking to the EM code to store return values from callling hooks, then filter them and based on the responses throw an email if more than 1 module tries to return a value for a hook like this.  But, REDCap doesn't honor the responses any more so what does this solve?

- Alternately, you can 'render' the message inside the hook and then, if you want to cancel (e.g. status = false, you can call an 'exit()' but this throws an EM error email...

So, we need to fix one or both of these things.  The quick and dirty fix is to render the code inside the hook and then call 'exit' for status-> false.

@taylorr4  - why was the behavior of this changed to ignore the return value of the hook?  There is even an example in the docs that suggest that you might be using this for Vanderbilt LDAP - is that not true?

-Andy

123andy:
I decided to bail on the return values from the hook and instead just added a method to gracefully exit in the event of a 'failed' lookup, e.g. `[ 'status'=>false ]` which seems to be working.

I did have work-around to return the values but it appears that the core REDCap code no longer handles them so it is irrelevant.  

For now, I'd like to get the pull request accepted.  Going forward, though, I think we should have a system where a hook could return a structured object that would drive REDCap, so making it possible to return values form an EM hook is important. 

Since hooks are evaluated across all modules, it is possible that many modules could have return values that would have to be handled by the core redcap code.  How to do this?

1) We could throw a warning if more than one module tries to control the 'result' object of a hook.  
2) We could return an array of return values, where each hook that has a return value is a member of the resulting array.  The problem here is that I dont know how REDCap core code would handle this.  In most cases, a return value is expected to be a single object.

So, I say that we capture the result.  Something like:

In startHook:
```
		$result = null;
		foreach ($hookNames as $thisHook) {
[...]

PR #164, opened 2019-02-23
Allow exit

123andy:
Adding ability to force exit from inside an EM hook - required to get redcap_custom_verify_username to work without the ability to pass return parameters.  This is also relevant to #89 for @lsgs and @mmcev106 .

Inside my config.json
```
  "allow-exit": [
    "redcap_cuxstom_verify_username"
  ],
```

Inside my EM hook:
```
public function redcap_custom_verify_username($username) {
    ....
    // Previously with this hook, you returned $status and $message, but now you have to render the results yourself.
    if ($status) {
        if (strlen($message) > 0) {
            echo "<div class='alert alert-success text-center'>" . $message . "</div>";
        }
    } else {
        echo "<div class='alert alert-danger text-center'>" . $message . "</div>";
        exit();  // This prevents the rendering of the rest of the user rights page...
    }
}
```

### DIFF
--- a/classes/ExternalModules.php
+++ b/classes/ExternalModules.php
@@ -94,6 +94,8 @@ class ExternalModules
 	
 	private static $deletedModules;
 
+	private static $allowExit;
+	
 	private static $configs = array();
 
 	# two reserved settings that are there for each project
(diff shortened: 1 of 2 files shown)


mmcev106:
Per past discussions, I'm not sure we can or should ever support exiting in the middle of a hook, or any other action that allows one module to change the behavior of another module without some kind of consent from the latter module.

In a few hours, I believe I could made the required changes in REDCap and the module framework to allow return codes to work with the **redcap_custom_verify_username** hook.   I think the *delayModuleExecution()* method might even be used to allow multiple modules to run **redcap_custom_verify_username** on the same project.  I believe this solution will be easier to support long term.  If this is holding you back, would you like me to see if I can get the OK to work on this?

mmcev106:
Per issue #89 we landed on a different strategy (at least for now).  Closing...

PR #165, opened 2019-02-26
Add a "button" setting type

mmcev106:


docs/official-documentation.md
```diff
@@ -119,6 +120,18 @@ Below is a *mostly* comprehensive list of all items that can be added to the  **
 	* **DEPRECATED (for now): Default values do NOT currently work consistently, and will likely need to be re-implemented.** Both project-settings and system-settings may have a **default** value provided (using the attribute "default"). This will set the value of a setting when the module is enabled either in the project or system, respectively.
 * If your JSON is not properly specified, an Exception will be thrown.
```
mmcev106: I'm not sure I understand how to use the example below.  Lately we've been implementing examples in the configuration example module included in the framework for clarity.  How would you feel about moving this there?

docs/official-documentation.md
```diff
@@ -119,6 +120,18 @@ Below is a *mostly* comprehensive list of all items that can be added to the  **
...
 
+#### Examples of button
+A basic case.
+
+``` json
+"url": {
+    "name": "Check information",
+    "value": "check.php?data=1",
+    "message":"The information has been checked"
```
mmcev106: As opposed to having a hard coded response value that is the same every time, would it make sense to display the response returned from the URL like we do in **validateSettings()**?  I think this would allow for much more flexibility, and encourage developers to implement post success and failure messages that are different.

manager/js/globals.js
```diff
@@ -457,6 +457,8 @@ ExternalModules.Settings.prototype.getColumnHtml = function(setting,value,classN
 
 			inputHtml += this.getInputElement(type, key, choice.value, inputAttributes) + '<label>' + choice.name + '</label><br>';
 		}
+	}else if(type == 'button'){
+         inputHtml = this.getButtonElement(type, key, setting.url.name,setting.url.value,setting.url.message, inputAttributes);
```
mmcev106: Would you mind adjusting the indentation so that it matches the surrounding lines?

manager/js/globals.js
```diff
@@ -593,6 +595,29 @@ ExternalModules.Settings.prototype.getFileFieldElement = function(name, value, i
...
+		}else{
+            $('[name ='+value+']').attr('href',url);
+            $('[name ='+value+']').attr('target','_blank');
+        }
+
+    });
+	return btn;
+
+}
+
```
mmcev106: I think there are some indentation discrepancies in the lines above too that are probably worth fixing.

docs/official-documentation.md
mmcev106: Eva agreed & moved the example.

docs/official-documentation.md
mmcev106: Eva agreed & implemented a json return object instead.

manager/js/globals.js
mmcev106: @knil-maloon, thanks for fixing these Eva!  I know they're not that important...

knil-maloon: You're welcome! They are also easy to fix ;)

PR #166, opened 2019-02-27
refactored to allow return values

123andy:
Here is an update to External Modules that supports return values.  It takes the approach that the last one wins.   Also, if there are return values from both EM hooks and 'non-em' function-based hooks, it takes the approach that the function-based hook wins.  This is debatable, but for now it is a start.  If we wanted we could try to allow multiple return values and handle the processing on the actual hook functions, but that seems like a leap when we don't have a good use case.

One issue that needs resolve is the exitAfterHook call.  Currently, if a hook has 'exitAfterHook' enabled, it happens before the return value is sent.  I'm guessing this may be okay as how can you process a return value AND exit after hook?  Seems a little hard.

If two EMs try to return a value for the same hook, it will throw a warning email to the admins, something like:

```
[This message was automatically generated by REDCap]

2 return values were generated from hook custom_verify_username by the following external modules:

1 admin_ninja: {"status":false,"0":"message","1":"SOMETHING ELSE"}
1 username_verification: {"status":false,"message":"The specified SUNet ID, <b><u>asdfsadfsa<\/u><\/b>, does not appear to be valid.<br\/><br\/> Many users have email aliases (e.g. Jane.Doe@stanford.edu) where the email prefix is not the same as their SUNet ID. A SUNet ID should be 8 characters or less without any periods or hyphens.<br\/><br\/>Try searching the <a href='https:\/\/stanford.rimeto.io\/search\/asdfsadfsa' target='_BLANK'> <div class='btn btn-xs btn-danger'><b>Stanford Directory<\/b><\/div><\/a> or contact your collaborator to obtain their SUNet ID."}
Only the last result from username_verification value will be returned to REDCap for processing by the hook function. Consider disabling or refactoring the other external modules so this does not occur.

URL: http://localhost/redcap_v8.9.3/UserRights/edit_user.php?pid=21
Server: localhost (2db2d4c7bc82)
User: andy123
```

It also requires an update to the `Hooks.php` class in REDCap.
```
	// Call a REDCap Hook function
	public static function call($function_name, $params=array())
	{
		// If not a real hook method in this class, then return
		if (!method_exists(__CLASS__, $function_name)) return false;

		// Call External Modules method before calling the native hook in the hook functions file
		if (defined("APP_PATH_EXTMOD")) {
			$em_result = \ExternalModules\ExternalModules::callHook($function_name, $params);
		}

		// Evaluate hook from hook_functions.php file if present
		if (function_exists($function_name)) {
			// Call the hook function
			$func_result = call_user_func_array($function_name, $params);
		}

		// We have to decide how to handle collisions where both the EM AND the hook function return a value
		// For now, I'll let the local function override the EM as any local hook function has to be
		// explicitly set up by the owner of the redcap instance
		$result = !empty($em_result)   && is_array($em_result)   ? $em_result   : false;
		$result = !empty($func_result) && is_array($func_result) ? $func_result : $result;

		// Call the appropriate method to process the return values, then return anything returned by the custom function
		return call_user_func_array(__CLASS__ . '::' . $function_name, array($result));
	}
```

While we are at it, I'd like to also do a minor update to the redcap_custom_verify_username hook:

```
	public static function redcap_custom_verify_username($result)
	{
		// If a message is returned, then output the message in a colored div
		if (isset($result['message']) && !empty($result['message'])) {
			// Set the color for the message based on the status
			$colorClass = $result['status'] ? 'green' : 'red';
			print RCView::div(array('class'=>$colorClass, 'style'=>'margin:10px 0;'), $result['message']);

			// If status is FALSE, then stop script execution
			if (isset($result['status']) && ($result['status'] === false)) exit;
		}
		// Don't return anything
	}
```

### DIFF
--- a/classes/ExternalModules.php
+++ b/classes/ExternalModules.php
@@ -1514,6 +1514,9 @@ class ExternalModules
 				self::$delayed[self::$hookBeingExecuted] = array();
 				self::$delayedLastRun = $lastRun;
 				foreach ($prevDelayed as $prefix=>$version) {
+					// Modules that call delayModuleExecution() normally just "return;" afterward, effectively returning null.
+					// However, they could potentially return a value after delaying, which would result in multiple entries in $resultsByPrefix for the same module.
+					// This could cause filterHookResults() to trigger unnecessary warning emails, but likely won't be an issue in practice.
 					$startHook($prefix, $version);
 				}
 			};
--- a/classes/ExternalModules.php
+++ b/classes/ExternalModules.php
@@ -1490,29 +1490,31 @@ class ExternalModules
 				$versionsByPrefix = self::getEnabledModules($pid);
 			}
 
-			foreach($versionsByPrefix as $prefix=>$version){
+			$startHook = function($prefix, $version) use ($arguments, &$resultsByPrefix){
 				$result = self::startHook($prefix, $version, $arguments);
-			    if (!empty($result) && is_array($result)) {
-                    // Lets preserve order of execution by order entered into the results array
-			        $resultsByPrefix[] = array(
-                        "prefix" => $prefix,
-                        "result" => $result
-                    );
-                }
+
+				// The following check assumes hook return values will always be arrays.
+				// That's totally fine for now since our only hook that returns a value does return an array.
+				// There may come a day when we'd want to support other types as well.
+				if (!empty($result) && is_array($result)) {
+					// Lets preserve order of execution by order entered into the results array
+					$resultsByPrefix[] = array(
+						"prefix" => $prefix,
+						"result" => $result
+					);
+				}
+			};
+
+			foreach($versionsByPrefix as $prefix=>$version){
+				$startHook($prefix, $version);
 			}
 
-			$callDelayedHooks = function($lastRun) use ($arguments){
+			$callDelayedHooks = function($lastRun) use ($startHook){
 				$prevDelayed = self::$delayed[self::$hookBeingExecuted];
 				self::$delayed[self::$hookBeingExecuted] = array();
 				self::$delayedLastRun = $lastRun;
 				foreach ($prevDelayed as $prefix=>$version) {
-					$result = self::startHook($prefix, $version, $arguments);
-    			    if (!empty($result) && is_array($result)) {
-    			        $resultsByPrefix[] = array(
-                            "prefix" => $prefix,
-                            "result" => $result
-                        );
-                    }
+					$startHook($prefix, $version);
 				}
 			};
 	
--- a/classes/ExternalModules.php
+++ b/classes/ExternalModules.php
@@ -96,8 +96,6 @@ class ExternalModules
 
 	private static $configs = array();
 
-	private static $returnValues = array();
-
 	# two reserved settings that are there for each project
 	# KEY_VERSION, if present, denotes that the project is enabled system-wide
 	# KEY_ENABLED is present when enabled for each project
--- a/classes/ExternalModules.php
+++ b/classes/ExternalModules.php
@@ -96,6 +96,8 @@ class ExternalModules
 
 	private static $configs = array();
 
+	private static $returnValues = array();
+
 	# two reserved settings that are there for each project
 	# KEY_VERSION, if present, denotes that the project is enabled system-wide
 	# KEY_ENABLED is present when enabled for each project
@@ -1342,10 +1344,18 @@ class ExternalModules
 		return "(" . implode(" OR ", $parts) . ")";
 	}
 
-	# begins the execution of a hook
-	# helper method
-	# should call callHook
-	private static function startHook($prefix, $version, $arguments) {	
+    /**
+     * begins execution of hook
+     * helper method
+     * should call callHook
+     *
+     * @param $prefix
+     * @param $version
+     * @param $arguments
+     * @return mixed|void|null  result from hook or null
+     * @throws Exception
+     */
+    private static function startHook($prefix, $version, $arguments) {
 		
 		// Get the hook's root name
 		if (substr(self::$hookBeingExecuted, 0, 5) == 'hook_') {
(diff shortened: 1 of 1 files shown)


classes/ExternalModules.php
```diff
@@ -1475,16 +1490,31 @@ static function callHook($name, $arguments, $prefix = null)
 				$versionsByPrefix = self::getEnabledModules($pid);
 			}
 
+			$startHook = function($prefix, $version) use ($arguments, &$resultsByPrefix){
+				$result = self::startHook($prefix, $version, $arguments);
+
+				// The following check assumes hook return values will always be arrays.
+				// That's totally fine for now since our only hook that returns a value does return an array.
+				// There may come a day when we'd want to support other types as well.
```
mmcev106: Just FYI, I extracted the duplicate lines below and added the comments above just for my own future reference.

classes/ExternalModules.php
```diff
@@ -1511,10 +1541,48 @@ static function callHook($name, $arguments, $prefix = null)
 			}
 		}
 
+        // As this is currently written, any function that returns a value cannot also exit.
+        // TODO: Should we move this to a shutdown function for this hook so we can return a value?
```
mmcev106: I'm on the same page as you that returning a value and calling exitAfterHook() would be an odd use case.  We can cross that bridge if/when we come to it.

classes/ExternalModules.php
```diff
@@ -1475,16 +1490,34 @@ static function callHook($name, $arguments, $prefix = null)
...
-			$callDelayedHooks = function($lastRun) use ($arguments){
+			$callDelayedHooks = function($lastRun) use ($startHook){
 				$prevDelayed = self::$delayed[self::$hookBeingExecuted];
 				self::$delayed[self::$hookBeingExecuted] = array();
 				self::$delayedLastRun = $lastRun;
 				foreach ($prevDelayed as $prefix=>$version) {
-					self::startHook($prefix, $version, $arguments);
+					// Modules that call delayModuleExecution() normally just "return;" afterward, effectively returning null.
+					// However, they could potentially return a value after delaying, which would result in multiple entries in $resultsByPrefix for the same module.
+					// This could cause filterHookResults() to trigger unnecessary warning emails, but likely won't be an issue in practice.
```
mmcev106: I added the above comments mainly for my own future reference.

classes/ExternalModules.php
```diff
@@ -1475,16 +1490,34 @@ static function callHook($name, $arguments, $prefix = null)
 				$versionsByPrefix = self::getEnabledModules($pid);
 			}
 
+			$startHook = function($prefix, $version) use ($arguments, &$resultsByPrefix){
+				$result = self::startHook($prefix, $version, $arguments);
+
+				// The following check assumes hook return values will always be arrays.
+				// That's totally fine for now since our only hook that returns a value does return an array.
+				// There may come a day when we'd want to support other types as well.
+				if (!empty($result) && is_array($result)) {
```
mmcev106: Just FYI, I extracted the duplicate lines below to this shared function so I could add the comments above for my own future reference.

classes/ExternalModules.php
```diff
@@ -1511,10 +1544,48 @@ static function callHook($name, $arguments, $prefix = null)
 			}
 		}
 
+        // As this is currently written, any function that returns a value cannot also exit.
+        // TODO: Should we move this to a shutdown function for this hook so we can return a value?
```
mmcev106: I'm on the same page as you that returning a value and calling exitAfterHook() would be an odd use case. We can cross that bridge if/when we come to it.

classes/ExternalModules.php
```diff
@@ -1511,10 +1544,48 @@ static function callHook($name, $arguments, $prefix = null)
...
 		if(self::$exitAfterHook){
 			exit();
 		}
-	}
+
+		// We must resolve cases where there are multiple return values.
+        // We can assume we only support a single return value (easier) or we can expand our definition of hooks
+        // to handle multiple return values as an array of values.  For now, let's shoot simple and just take
+        // the latest one and throw a warning to the admin
+		return self::filterHookResults($resultsByPrefix, $name);
```
mmcev106: I think just returning the last value is good starting point too. Modules would still be able to control execution order and as a result the return value via delayModuleExecution().  We can consider more complex solutions if/when we encounter a need.

classes/ExternalModules.php
```diff
@@ -1404,6 +1414,8 @@ private static function startHook($prefix, $version, $arguments) {
 		}
 
 		$instance->setRecordId(null);
+
+        return $result;
```
mmcev106: This is not a big deal to me, but there are a handful of tab vs. space discrepancies that I figured I should at least let you know about in case you wanted to normalize them. Plenty of those already exist in this project... At some point I am hoping come up with a reasonable way to enforce standards, then do a sweep of the code to clean everything up at once.  I'm going to go ahead and merge this as is.

123andy:
thx!

classes/ExternalModules.php
123andy: We use spaces internally but it seems Rob likes tabs most of the time...  I'm open to adopting any standard.  It would also be nice to have a doc about variable name, method names, camel case, underscores, etc...   Thanks for your work on this.

mmcev106:
I just sent a REDCap core PR request to Rob.  I decided to make some tweaks to the `call()` function before sending it over, the biggest one being that I changed it back to returning false when there are no return values (rather than passing false to `call_user_func_array()`).  I don't think that changes anything for your needs:

```
// Call a REDCap Hook function
public static function call($function_name, $params=array())
{
	// If not a real hook method in this class, then return
	if (!method_exists(__CLASS__, $function_name)) return false;

	// Call External Modules method before calling the native hook in the hook functions file
	if (defined("APP_PATH_EXTMOD")) {
		$result = \ExternalModules\ExternalModules::callHook($function_name, $params);
	}

	// Evaluate hook from hook_functions.php file if present
	if (function_exists($function_name)) {
		// Call the hook function
		$func_result = call_user_func_array($function_name, $params);
		if($func_result){
			// We have to decide how to handle collisions where both the EM AND the hook function return a value
			// For now, I'll let the local function override the EM as any local hook function has to be
			// explicitly set up by the owner of the redcap instance
			$result = $func_result;
		}
	}

	// If an array is not returned, then return
	if (!is_array($result)) return false;

	// Call the appropriate method to process the return values, then return anything returned by the custom function
	return call_user_func_array(__CLASS__ . '::' . $function_name, array($result));
}
```

123andy:
Sounds fine - no need to call the function if there aren't any parameters (at least so far).

-A

PR #178, opened 2019-05-08
Language (Translation) feature for the EM Framework

grezniczek:
## Hey guys,

So, I did it ... as briefly discussed with Mark [over at the community site](https://community.projectredcap.org/questions/45861/how-to-do-internationalization-and-localization-in.html?childToView=62740#comment-62740). I have added support for INI-Style language files for External Modules, similar to how REDCap core does it.
The implementation is totally transparant for existing modules. To use the new features, they have to opt. Strings ("name", "description" fields) can be translated without breaking backward compatibility. To use the language feature in module code (via a tt() function) will require the EM framework version 2 and REDCap >= where the language feature is implemented (so 9.0.1 the earliest ;).

## How does it work? 
- Add a "lang" folder with translation files (such as English.ini) into the EM's folder
- To translate config stuff, add "tt_name" or "tt_description" fields containing the key from the language file.
- To translate text in module code, use the tt() function provided by the framework version 2 (I added a forwarding from AbstractExternalModule via __call). Call as tt($key) or tt($key, $value1, $value2, ...) or tt($key, $valueArray) to get string interpolation, e.g. when the text corresponding to the key is e.g. "{0} are {1}." and $values = array("Apples", "red") this will give "Apples are red.". tt($key) without values would just pass through the original string (nice when interpolation happens in the browser).

## What is missing? 
- A corresponding JS function replicating tt for interpolation in the browser. I will work on this asap.

## How does this look in a module?
- Head over to the [nedCAPTCHA](https://github.com/grezniczek/redcap_nedcaptcha/tree/langfeature) external module and have a look, focus on the lang directory and on config.json. Use of tt() is made in NEDCaptchaExternalModule.php around line 600. Complete English and German translations are available (I should have picked another module to showcase this ... there was a LOT to translate).
- Install it with the update EM framework to see it in action.

*Let me know what you think!*

Cheers,
GR

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 /.idea
 /vendor
+/.vscode
--- a/classes/AbstractExternalModule.php
+++ b/classes/AbstractExternalModule.php
@@ -1257,6 +1257,11 @@ class AbstractExternalModule
 			return call_user_func_array([$this, 'log_internal'], $arguments);
 		}
 
+		// Forward translation support function to the framework.
+		if ($name == "tt" && isset($this->framework) && method_exists($this->framework, $name)) {
+			return call_user_func_array([$this->framework, $name], $arguments);
+		}
+
 		throw new Exception("The following method does not exist: $name()");
 	}
 
--- a/classes/ExternalModules.php
+++ b/classes/ExternalModules.php
@@ -37,6 +37,39 @@ class ExternalModules
 	const KEY_ENABLED = 'enabled';
 	const KEY_DISCOVERABLE = 'discoverable-in-project';
 	const KEY_CONFIG_USER_PERMISSION = 'config-require-user-permission';
+	
+	//region Language feature-related constants
+
+	/**
+	 * The name of the system-level language setting.
+	 */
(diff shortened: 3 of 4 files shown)


mmcev106:
This looks fantastic.  Thanks very much for contributing this @grezniczek!  I have not had time to do any work on growing the framework for the past month, but am hoping to get caught up within the next couple of weeks and review this (really just for long term support concerns).

grezniczek:
Hello @mmcev106, good to hear :)
I just did some more work towards localization of the EM Framework itself, i.e. extracting translateable strings. All PHP stuff is done. JS is still to do (need to get the JS-version of tt working -- also, probably need to think a bit more about sanitation and maybe revisit a few places).
But the strings are there, in English.ini, which ultimately should get merged with REDCaps main language file (for now, I am appending them to global $lang when the EM framework is initialized).

I will open up a new pull request from a different branch I created (probably should have done this from the start) - this will include all changes, so probably you can get just ignore this pull request (not a GitHub expert, but I am learning).

I also found a few minor bugs, one was an implode() with parameters reversed leading to only show first module author on the system manager.

PR #180, opened 2019-05-16
Fix module link icons not working any more after upgrade to REDCap 9.0.1

grezniczek:
REDCap 9.0.1 brought FontAwsome icons to project and control center menu items.
This broke EM links on project pages, as the strategy was to clone an existing link and replace the src attribute of the img tag. Now with FontAwsome icons there of course is no img anymore, so this fails.
All EM links had bells.

This PR fixes this and adds the ability to use FontAwsome icons for EM links.

PS: Please disregard the first 3 commits ... I started to commiting to testing some time ago instead of creating a new branch ... hence the "restore to Vandy state".


mmcev106:
I apologize, I did not get caught up on email before I starting working to fix this issue myself.  Here's the fix I came up with to support existing icons:

https://github.com/vanderbilt/redcap-external-modules/commit/3f85d35daa30c6059b556b8c8f83cd08348a174b

I'm also working on a change to require font-awesome icons be specified instead of png filenames when authors update their modules to framework version 3.  This change would encourage standardization over time, and I think it'd be a simpler change all around.  I'm thinking perhaps we should close this pull request.  However, would you like to make a case for an alternate approach before I do?

grezniczek:
Well, it's up to you, @mmcev106

The fix / changes I made would allow the use FA icons as soon as the fix makes it into production without breaking existing png icons. 
Maybe the docs could be further updated to recommend module authors who choose to use FA icons with framework version 2, to also include a png with the same name in the EM's images folder for backward compatiblity with earlier REDCap versions.

Is version 3 on the table soon? What will it change? If it's only API additions (like the i18n API), I would stay with version 2, as there are no breaking changes.

mmcev106:
I think I'm going to go a different direction and make the font-awesome icon switch a breaking change for version 3.  I think it will be easier to support long term, which is my main goal these days since time such a constraint.

I guess I'll go ahead and close this.  Thanks VERY much for the contribution though.  It is much appreciated.

PR #185, opened 2019-06-27
Any auth

ihabzee:
allow any authenticated user to access ANYAUTH pages without project privileges


mmcev106:
@ihabzee, @123andy,

I've done a basic review of this and read Andy's description here:

https://github.com/vanderbilt/REDCap/pull/136

Please forgive me, I'm guessing I'm missing something, but I want to make sure.  Can the same goal not already be accomplished using system level module pages?  They can have authentication, and could still be made project specific by specifying the project id as something other that the "pid" GET param.

Also, are the branching logic changes in this PR related to ANYAUTH, or just general branching logic changes?

Thanks,
Mark

123andy:
Hi Mark,

1) @ihabzee is going to remove the last commit - it is part of a separate feature fix/addition.
2) For your question - there are cases where you want to have a 'project-specific' page so a single EM can be reused by many projects, each of which has its own 'instance' of the EM and its own 'instance' of permissions / users / etc.   It is for this case that ANYAUTH comes in handy.   One example we are working on is a scheduling module where you create appointments in a redcap project (as records).  The EM is enabled for each 'calendar' of events and it creates project-level pages that allow people to book appointments.  In some cases, you may want to force authentication for such a page, but not in the current format of 'current project users' - but rather in the format of ANY project user.  It is the equivalent of the behavior that was possible with plugin pages and the allowUsers() even when used in a project-level context...

Somewhat related, this is further complicated by shibboleth...  We could hack around this by using noauth and relying on shibboleth to do the auth for us using the normal/api urls, but this would mean our module would not work for other REDCap installs that use table or ldap/table.  So, this is a work-around for that...

We hope to integrate simpleSAML in the future, which might make it easier for EM developers to play nice with both SAML and table/ldap installs...

ihabzee:
@123andy @mmcev106 I removed commit [76883ae](https://github.com/vanderbilt/redcap-external-modules/pull/185/commits/76883ae3830cf917500828b425064f655600c5ef ) and created new branch for [Children Logical  branching](https://github.com/susom/redcap-external-modules/tree/children-branching-logic) which still I need to review with Andy before creating a pull request thanks,

index.php
```diff
@@ -72,10 +86,12 @@
...
-	$link = $module->redcap_module_link_check_display($_GET['pid'], $link);
-	if (!$link) {
-		die("You do not have permission to access this page.");
-	}
+    if (!ANYAUTH) {
+        $link = $module->redcap_module_link_check_display($_GET['pid'], $link);
+        if (!$link) {
+            die("You do not have permission to access this page.");
+        }
+    }
```
mmcev106: I don't think we want ANYAUTH to bypass `redcap_module_link_check_display` because it forces a single use case and takes link visibility control away from each module.  I think it might be more appropriate update the default implementation for this method in `AbstractExternalModules` to check ANYAUTH and change it's return value accordingly.

mmcev106:
This change seems like bad design to me...  However, I'm willing to trust you @123andy and assume we're just thinking about different use cases.  Apart from my `redcap_module_link_check_display` comment, I'm ready to merge this.  

If you're interested in further discussion, it seems like this should somehow be rolled into existing functionality.  Making project level calls from a system context is already a common use case today.  Ideally all project functionality should still be available from a system context.  I know `allowsUsers()` and the get/set project setting modules methods currently work from a system module page.  If there's some project function that can't be used from a system context currently, I think fixing that might be more appropriate/useful than adding an additional authentication method and a pseudo-project context. 
 I've thought about adding a `setProjectId()` method to the framework to allow modules to switch between multiple project contexts within the same request/process (would be very useful in crons).  Feel free to call me at (615) 343-5350 if that would be easier.

123andy:
I'm on vacation but will flag for follow-up.  We can chat once I'm back and come up with the best solution...  thx!

mmcev106:
@123andy, have you had any more thoughts on this?  I'm still happy to just merge it if you think it's just a different use case and my concerns aren't quite relevant.

123andy:
Mark,

We've made an attempt to refactor our code which involves:

using the link in config.json for a project where the EM is enabled.  When the link is clicked, our php code removes the &pid=xxx and replaces it with &project_id=xxx and redirects the client to a system-level page (so project-level permissions aren't enforced).  We then take the $_GET['project_id'] to create a new instances of \Project and try to use methods from there (As not all plugin methods currently support a project_id level override).  It seems to be working so, for now, we can work-around the any-auth.

I still think the 'concept' is valid - it should be possible to create project-level EM that has a url that is 'open' to any REDCap user (vs. ANY -- NOAUTH user) but, for now, it looks like we may be able to work-around...

-Andy

mmcev106:
I'm glad you found a workaround.  I'd love to keep trying to understand the concept from your angle (now or in the future).  My only hesitation to merging this is still just the `redcap_module_link_check_display` comment on the review.

mmcev106:
I'll close this since it hasn't been updated in a couple of months.  Feel free to reopen it anytime.

PR #187, opened 2019-07-10
small fix for getSubSettings() function inside the Absct…

123andy:
Small fix for oversight in getSubSettings() function inside the AbstractExternalModule class. Previously only called subsettings for project settings, and not system settings.   Changed the function from private to static for getSettingsAsArray in ExternalModules class to allow for this change

### DIFF
--- a/classes/AbstractExternalModule.php
+++ b/classes/AbstractExternalModule.php
@@ -269,8 +269,7 @@ class AbstractExternalModule
 		foreach($config['sub_settings'] as $subSetting){
 			$keys[] = $this->prefixSettingKey($subSetting['key']);
 		}
-
-		$rawSettings = ExternalModules::getProjectSettingsAsArray($this->PREFIX, self::requireProjectId($pid));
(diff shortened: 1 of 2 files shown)


mmcev106:
Thank you @123andy.  I went ahead and merged this.  However, I should make sure you know that this method is deprecated because it has some limitations with complex nested subsettings.  The `framework->getSubSettings()` method fixes this, but isn't quite backward compatible.  We likely need to make this same change there...  See the documentation for each framework version for more details:

https://github.com/vanderbilt/redcap-external-modules/blob/testing/docs/framework/intro.md

mmcev106:
I'm sorry @123andy, I reviewed this a little too hastily.  I had to revert it in commit b351a29 because it broke a unit test.  The current behavior depends on project id detection in `requireProjectId()`.  It seems like `detectProjectId()` would be more appropriate here, but even it has a slightly different behavior since it doesn't account for the `PROJECT_ID` constant.  I think this is going to require a little more refactoring to ensure it doesn't break any modules.

123andy:
Thanks mark - I'll refer it back to the developer and find a better fix.

PR #190, opened 2019-07-29
Subsettings fix

irvins:
New Function specifically for getting System Sub Settings.  Should not conflict as the previous pull request did.


mmcev106:
@irvins, thanks very much for the PR!  The `$module->getSubSettings()` method is actually deprecated in favor of `$module->framework->getSubSettings()`, so we would likely want to make any new changes in the latest framework version class.  We now add all new methods there to prevent conflicts with existing modules.  Would you mind looking into implementing a similar solution there?

mmcev106:
I guess I'll close this since I haven't heard back in about three months.  Feel free to reopen it anytime.

PR #195, opened 2019-08-13
Timed cron related changes

mmcev106:
@scottjpearson, this PR aims to take care of the immediate concerns with timed crons so that the testing branch can move to production without having to revert the timed cron changes.  It includes the following changes:

- Previously, long running cron detection started sending emails when the next cron started after the previous cron job had run for at least an hour.  This delay period has been bumped up to 23 hours (as opposed to 24, see the large comment in the code on why).  Also, a limit has been added to make sure only one email is sent per 23 hour period.  This improves the case where a cron scheduled for every minute takes more than 23 hours to run (previously it would start sending emails every minute).
- The timed cron documentation has been removed since we have a few concerns to address before publicly supporting this feature.  I'll create an issue shortly with these details and instructions for re-adding this documentation.

What are your thoughts on the long running cron detection changes?  Do you have any other concerns about the timed cron changes in the testing branch moving to production, and the consortium shortly afterward?  They'd be undocumented and likely not used, so I'm mainly thinking in terms of changing existing functionality.  Thanks!

### DIFF
--- a/classes/ExternalModules.php
+++ b/classes/ExternalModules.php
@@ -39,6 +39,7 @@ class ExternalModules
 	const KEY_CONFIG_USER_PERMISSION = 'config-require-user-permission';
 
 	const KEY_RESERVED_IS_CRON_RUNNING = 'reserved-is-cron-running';
+	const KEY_RESERVED_LAST_LONG_RUNNING_CRON_NOTIFICATION_TIME = 'reserved-last-long-running-cron-notification-time';
 
 	const TEST_MODULE_PREFIX = 'UNIT-TESTING-PREFIX';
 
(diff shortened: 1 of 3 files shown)


mmcev106:
I believe the call to remove the `KEY_RESERVED_LAST_LONG_RUNNING_CRON_NOTIFICATION_TIME` setting in the `lockCron()` method should take care of cases 1 & 2.  It effectively resets the setting each time a cron successfully starts.  What do you think?

scottjpearson:
If that's the case, I'm ok with things as they stand.

mmcev106:
@scottjpearson, thanks very much for the review!

PR #198, opened 2019-08-27
Cron

scottjpearson:
Mark,

I think we're finally ready for your review.

Thanks!
Scott


classes/AbstractExternalModule.php
```diff
@@ -1630,4 +1631,36 @@ public static function init()
...
+		$modifications = $this->getModifiedCrons();
+		if ($modifications) {
+			return $modifications;
+		}
+		$config = $this->getConfig();
+		if (isset($config['crons'])) {
+			return $config['crons'];
+		}
+		return array();
+	}
```
mmcev106: I think we might want to move these new `AbstractExternalModule` methods to the `ExternalModules` class instead for a few reasons:

- We do not add methods or variables to the `AbstractExternalModule` anymore since it has caused conflicts with existing modules int the past.  These methods could go in the latest [framework ](https://github.com/vanderbilt/redcap-external-modules/blob/testing/docs/framework/intro.md) class if `ExternalModules` isn't the most appropriate spot.
- It would prevent module authors from overriding them.
- It prevents a conflict with the latest version of the testing branch (and would hopefully allow a clean merge afterward).

classes/AbstractExternalModule.php
```diff
@@ -1630,4 +1631,36 @@ public static function init()
 		self::$RESERVED_LOG_PARAMETER_NAMES_FLIPPED = array_flip(self::$RESERVED_LOG_PARAMETER_NAMES);
 		self::$LOG_PARAMETERS_ON_MAIN_TABLE = array_flip(array_merge(self::$RESERVED_LOG_PARAMETER_NAMES, self::$OVERRIDABLE_LOG_PARAMETERS_ON_MAIN_TABLE));
 	}
+
+	public function removeModifiedCrons() {
+		if ($this->getModifiedCrons()) {
```
mmcev106: Not that it really matters, but this `if` check is redundant since the `removeSystemSetting()` already performs this check internally.

classes/AbstractExternalModule.php
```diff
@@ -1630,4 +1631,36 @@ public static function init()
...
+	public function setModifiedCrons($cronSchedule) {
+		foreach ($cronSchedule as $cronAttr) {
+			if (!ExternalModules::isValidCron($cronAttr)) {
+				throw new \Exception("The following cron is not valid! ".json_encode($cronAttr));
+			}
+		}
+		$this->setSystemSetting(self::$RESERVED_CRON_MODIFICATION_NAME, $cronSchedule);
+	}
+
+	public function getCronSchedules() {
```
mmcev106: Does this method work properly on module updates?  Here's an example scenario:

- Modified crons are stored
- A module update is installed that adds a new cron or changes the method called for an existing cron
- Will these new crons info from the new module version be overridden by the modified crons in the db?

If so, we may want to store only the time information instead of the entire cron config.

scottjpearson: So upon an module update:
- modified crons will continue to be modified (overridden). This seems to me to be the correct behavior. Otherwise, you'd have to fine tune the system after each upgrade.
- crons that have been removed will disappear (while their modifications will remain in the saved array - a very minor but present issue - tolerable).
- new crons will have no override in place; therefore, their default times will be used.
- the interface can be used to adjust these.

scottjpearson:
Changes made. All clear, from my end, to merge.

classes/AbstractExternalModule.php
mmcev106: This might be easier to talk through in person on Tuesday.  Looking at the code, I don't believe the behavior you described is what would actually happen.  In `getCronSchedules()`, it looks like everything in `$config['crons']` will be ignored once `getModifiedCrons()` starts returning value.  I'm still thinking we might have to only override the time fields instead of the entire cron configuration.

scottjpearson:
As discussed, the changes have been made. Also, code for the unit tests has been changed, but I have thus far been unable to get them to run. Please run these as well.

classes/AbstractExternalModule.php
mmcev106: The latest commit fixes this :)

mmcev106:
@scottjpearson, I'm sorry, I believe there are still a few unresolved conversations above, and the `testCheckCronModifications()` test is failing currently.  Let me know if I can help with getting tests running locally.

classes/AbstractExternalModule.php
scottjpearson: Again, I tend towards defensive programming over unnecessary optimization as a rule. (In other words, the biggest risk to our code is a future modification that accidentally violates this rule than running a few extra CPU cycles.) So I'm leaving this as is.

classes/AbstractExternalModule.php
```diff
@@ -24,6 +24,7 @@ class AbstractExternalModule
 	private static $RESERVED_LOG_PARAMETER_NAMES_FLIPPED;
 	public static $OVERRIDABLE_LOG_PARAMETERS_ON_MAIN_TABLE = ['timestamp', 'username', 'ip', 'project_id', 'record', 'message'];
 	private static $LOG_PARAMETERS_ON_MAIN_TABLE;
+	public static $RESERVED_CRON_MODIFICATION_NAME = "___RESERVED___MODIFIED_CRON___";
```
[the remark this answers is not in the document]
scottjpearson: Done

classes/ExternalModules.php
```diff
@@ -4117,4 +4151,57 @@ public function getUnsafeEDocReferences()
...
+					throw new \Exception("The following cron is not valid! ".json_encode($cronAttr));
+				}
+			}
+			$module->setSystemSetting(self::$RESERVED_CRON_MODIFICATION_NAME, $cronSchedule);
+		} else {
+			throw new \Exception("Could not instantiate module '$modulePrefix'!");
+		}
+	}
+
+	public static function getCronSchedules($modulePrefix) {
```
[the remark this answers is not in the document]
scottjpearson: Please do make the changes as you see fit.

mmcev106:
@scottjpearson, I just updated PHPUnit to the latest version that supports PHP 5.6.  Do you still see that "must be compatible" PHPUnit error after doing the following?
- Do a `git pull` on the `cron`
- Run `composer install`
- Run `./run-tests.sh`

mmcev106:
@scottjpearson, if that doesn't work, the following might:
- Run `composer global require phpunit/phpunit=6.*` (the latest composer version that supports the base test class used in 5.x)
- Instead of running `./run_tests.sh`, run ` ~/.composer/vendor/phpunit/phpunit/phpunit`

classes/ExternalModules.php
```diff
@@ -4156,7 +4157,7 @@ public static function removeModifiedCrons($modulePrefix) {
 		$module = self::getModuleInstance($modulePrefix);
 		if ($module) {
 			if ($module->getModifiedCrons($modulePrefix)) {
-				$module->removeSystemSetting(self::$RESERVED_CRON_MODIFICATION_NAME);
+				$module->removeSystemSetting(KEY_RESERVED_CRON_MODIFICATION_NAME);
```
mmcev106: I think the `self::` reference here and the other two places in this file below might still be required (just without the dollar sign).  I don't believe the const is actually being used currently.

mmcev106: Please disregard this.  I'll end up editing those lines anyway per this comment:

https://github.com/vanderbilt/redcap-external-modules/pull/198/files#r319550342

classes/AbstractExternalModule.php
mmcev106: My reasoning for this is not actually to preserve CPU cycles, but rather to avoid code suggesting cases/possibilities exist that should not, and to [fail-faster] if they do.

classes/ExternalModules.php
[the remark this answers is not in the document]
mmcev106: Done

classes/AbstractExternalModule.php
scottjpearson: You're the maintainer of the code, so you should get deference. (My way makes more sense to me, but I think we're in an endless struggle.) It's fine that you made the change. Thanks.

mmcev106:
Scott & I already talked about this, but I'll keep this thread up to date for my own reference.  I fixed a couple unit tests that were only failing on Scott's PC.  There is one more test failing that I'd like to troubleshoot on his PC, but he can run most tests currently.

scottjpearson:
All unit tests are passing for me. Please merge @mmcev106

mmcev106:
@scottjpearson, merged!  Per our normal deploy schedule these changes should move to production on the 23rd.  Let me know if it's worth looking into moving these to production sooner.

scottjpearson:
23rd is fine. Do we know yet what minimum version of REDCap this will be in?

scottjpearson:
And THANKS!

mmcev106:
You're welcome!  It's hard to know for sure what version a feature will be included since we have unexpected bug releases sometimes...  I've been scheming ways to automatically detect the version for documented methods after the fact (not sure when that will happen).

PR #204, opened 2019-09-19
Pass project ID in setDag and setData 

ihabzee:
In some use cases we wanted to set DAG from different project. I added optional `$projectId` parameter to `setDag` and `setData` 
@123andy


123andy:
Ihab - I still this this is a good contribution.   Since it doesn’t break anything I’m not sure if it should be part of v2 or v1?

Issue #211, opened 2019-10-21
Document the REDCap version numbers for each framework version


pbchase:
It would be good to know what REDCap version number is required to get each Framework version. This is similar to Issue #201, but less specific.

mmcev106:
Our release process makes this a little difficult, since it's unclear what version a features will make it into at commit time.  However, I think we could write a script to fill in versions after the fact (perhaps on a weekly basis).  This script could also list the REDCap version in which each method first appeared for each framework version

PR #212, opened 2019-10-21
Add REDCap version details to framework version docs

pbchase:
This PR addresses Issues #211. I added the REDCap version numbers on the intro.md and the docs specific to each framework. Where to place the new version details is in each is rather subjective. Let me know what you think. 

Also, I bend the truth a bit in intro.md when I say the Framework Version _requires_ a certain REDCap version. I know that strictly speaking one can load newer EM code on older REDCap versions, but do we think that ever happens in a production environment? I chose brevity and being probably right over the verbosity required to be 100% accurate.

### DIFF
--- a/docs/framework/intro.md
+++ b/docs/framework/intro.md
@@ -1,4 +1,7 @@
 ## External Module Framework - Versioning Introduction
+
+### How to specify the Framework Version for a module
+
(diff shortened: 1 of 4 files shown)


mmcev106:
Thanks very much @pbchase!

PR #225, opened 2019-11-19
Fix an incompatibility with IE11 in the Javascript Module Object's interpolate function

grezniczek:
IE11 doesn't like regexp literals. Using the "old" way of creating the RegExp fixes the problem.

See issue #226


mmcev106:
I also went ahead and merged this into our `production` branch, and made a note to include it in the merge to the `release` branch next week so the initial language release to the consortium includes it.

grezniczek:
Great! Thanks!

PR #232, opened 2019-12-13
Feature to allow user's with design/setup rights to enable modules on their own

taylorr4:
… their own


mmcev106:
Closing after cherry-picking the commit into PR #234 against the release branch.  There are changes in the testing & production that we likely don't want to merge to the release branch just yet.

Issue #194, opened 2019-08-08
Change the repo to use the latest description from config.json


mmcev106:
It currently uses the description manually entered with the first version submissions, which I think we can get rid of in favor of the description in config.json.

mmcev106:
Closing as a duplicate of #200

PR #234, opened 2019-12-17
Additions to the feature to allow users request a module to be enabled

mmcev106:


classes/ExternalModules.php
```diff
@@ -1682,8 +1701,13 @@ private static function setSetting($moduleDirectoryPrefix, $projectId, $key, $va
...
 		};
 
+		// If module is being enabled for a project and users can activate this module on their own, then skip the user-based permissions check
+        // (Not sure if this is the best insertion point for this check, but it works well enough.)
+		$skipUserBasedPermissionsCheck = ($key == self::KEY_ENABLED && is_numeric($projectId)
+                && ExternalModules::getSystemSetting($moduleDirectoryPrefix, ExternalModules::KEY_USER_ACTIVATE_PERMISSION) == true && ExternalModules::hasDesignRights());
+
 		try{
-			if (self::areSettingPermissionsUserBased($moduleDirectoryPrefix, $key)) {
+			if (!$skipUserBasedPermissionsCheck && self::areSettingPermissionsUserBased($moduleDirectoryPrefix, $key)) {
```
mmcev106: I'm not coming up with an easier way to do the `$skipUserBasedPermissionsCheck` check.  It just seems like a surprisingly complicated case with the way module settings/permissions are currently designed.

classes/ExternalModules.php
```diff
@@ -5091,4 +5115,29 @@ public static function getCronSchedules($modulePrefix) {
...
+				$project_url = APP_URL_EXTMOD . 'manager/project.php?pid=' . $project_id;
+
+				$from = $project_contact_email;
+				$fromName = $project_contact_name;
+				$to = [$userInfo['user_email']];
+				$subject = "[REDCap] External Module \"{$config['name']}\" has been activated";
+				$message = "The External Module \"<b>{$config['name']}</b>\" has been successfully activated for the project named \""
+					. \RCView::a(array('href' => $project_url), strip_tags($app_title)) . "\".";
+				$email = self::sendBasicEmail($from, $to, $subject, $message, $fromName);
+			} catch (Exception $e) { }
```
mmcev106: Why are we ignoring this exception?

manager/ajax/enable-module.php
```diff
@@ -1,14 +1,21 @@
...
+
+use Exception;
+
 require_once dirname(dirname(dirname(__FILE__))) . '/classes/ExternalModules.php';
 
-// Only administrators can enable/disable modules
-if (!SUPER_USER) exit;
+// Can the current user enable/disable modules?
+$userCanEnable = (SUPER_USER || (ExternalModules::hasDesignRights() && ExternalModules::getSystemSetting($_POST['prefix'], ExternalModules::KEY_USER_ACTIVATE_PERMISSION) == true));
+if (!$userCanEnable) exit;
```
mmcev106: These lines look (logically) identical to the ones added to `disable-module.php`.  Should we extract them to a new function on the `ExternalModules` class to avoid duplicating this logic?

manager/ajax/get-disabled-modules.php
```diff
@@ -103,6 +103,7 @@
 			$enabled = ExternalModules::getProjectSetting($prefix, $_GET['pid'], ExternalModules::KEY_ENABLED);
 			$system_enabled = ExternalModules::getSystemSetting($prefix, ExternalModules::KEY_ENABLED);
 			$isDiscoverable = (ExternalModules::getSystemSetting($prefix, ExternalModules::KEY_DISCOVERABLE) == true);
+			$userCanEnable = (SUPER_USER || (ExternalModules::hasDesignRights() && ExternalModules::getSystemSetting($prefix, ExternalModules::KEY_USER_ACTIVATE_PERMISSION) == true));
```
mmcev106: This looks like it duplicates the aforementioned checks as well.

manager/templates/enabled-modules.php
```diff
@@ -260,6 +260,7 @@
 			$enabled = false;
 			$system_enabled = ExternalModules::getSystemSetting($prefix, ExternalModules::KEY_ENABLED);
 			$isDiscoverable = (ExternalModules::getSystemSetting($prefix, ExternalModules::KEY_DISCOVERABLE) == true);
+			$userCanEnable = (SUPER_USER || (ExternalModules::hasDesignRights() && ExternalModules::getSystemSetting($prefix, ExternalModules::KEY_USER_ACTIVATE_PERMISSION) == true));
```
mmcev106: This looks like it duplicates the aforementioned checks as well.

manager/ajax/disable-module.php
```diff
@@ -2,8 +2,9 @@
 namespace ExternalModules;
 require_once dirname(dirname(dirname(__FILE__))) . '/classes/ExternalModules.php';
 
-// Only administrators can enable/disable modules
-if (!SUPER_USER) exit;
+// Can the current user enable/disable modules?
+$userCanDisable = (SUPER_USER || (ExternalModules::hasDesignRights() && ExternalModules::getSystemSetting($_POST['module'], ExternalModules::KEY_USER_ACTIVATE_PERMISSION) == true));
```
mmcev106: This looks like it duplicates the aforementioned checks as well.

mmcev106:
I can finish this up if you're struggling on time more than I am.

classes/ExternalModules.php
taylorr4: @mmcev106 not sure, really. Maybe i wasn't sure what we'd normally return in this kind of exception. Do with it what you will.

manager/ajax/enable-module.php
taylorr4: True. It might be worth doing. Or it could always be done later.

manager/ajax/disable-module.php
taylorr4: true

taylorr4:
@mmcev106 okay, i abstracted the $userCanEnable logic into a new method

taylorr4:
@mmcev106 Okay, i think it's more or less ready to merge. What do you think?

classes/ExternalModules.php
taylorr4: Okay, i added something hear to deal with the exception.

mmcev106:
I've merged this to the **release** branch, and created PR #194 for REDCap core (for the consortium).  I also did do some basic smoke testing.  I don't anticipate any issues, but you may want to test the new functionality again from REDCap core since this skipped our normal merge & testing procedure/timeline.

Issue #236, opened 2019-12-18
version redirect for EM URLs


123andy:
I think we have an issue where the 'redirect' to the latest version of REDCap doesn't work for External Module urls.

In this example, the current server is 9.5.1 but the url was bookmarked as 9.3.5

So, if someone has:
https://you.edu/redcap_v9.3.5/ExternalModules/?...

The redirect tries to do so with an include of the  9.5.1 External modules but since the class has already been loaded from the original old version page hit, this results in an error (redeclaring the same object).

I think the answer is probably to have the source code do a redirect to the current EM version if there is an older version detected in an EM url but I haven't had the time to look deeply.   I think the code is in the actual REDCap bootstrap and not EM framework... (@taylorr4 )

grezniczek:
I cannot see this behavior. Trying with the MySQL Simple Admin, calling a 9.3.5-versioned link to the plugin page in a 9.5.1 REDCap, this works as expected. The switch to the proper version is made with the call to `ExternalModules::callRedcapConnect();` (which calls the unversioned redcap_connect.php) in ExternalModules.php. 
While both version (from the link and current) of ExternalModules and AbstractExternalModule classes are included, I don't see any re-declaration error (at least not on PHP7.3.3), and I could confirm that only the current versions of these are actually used by any subsequent code.
So the redeclaring error may be PHP version specific?

However, it might be a good idea to move the version detection to an earlier time point, before the EM classes are included.

```
$noAuth = isset($_GET['NOAUTH']);
if($noAuth){
    // This must be defined at the top before redcap_connect.php is required.
    define('NOAUTH', true);
}

if(!defined('APP_PATH_WEBROOT')){
    if(!defined('PLUGIN')){
        // Since a change to redcap_connect.php on 4/6/18, this is required to make sure REDCap is initialized for command line calls like cron jobs.
        define('PLUGIN', true);
    }
    $connectPath = dirname(dirname(__DIR__)) . DIRECTORY_SEPARATOR . "redcap_connect.php";
    require_once $connectPath;
}
```
at the top of `ExternalModules/index.php` should work.

This would ensure that a minimal amount of "old" code is loaded/executed (in this case only the code shown above).

grezniczek:
Hmm, this indeed appears to depend on the PHP version. On a production system with 7.0.33, I get blank pages when calling the MySQL Simple Admin module with an outdated version link. So I have to assume the error described by @123andy occured.

mmcev106:
[...]

PR #237, opened 2019-12-19
Fixed a case where redirections from old REDCap versions caused class…

mmcev106:
… redeclaration errors in some scenarios.

This aims to fix issue #236.


taylorr4:
@mmcev106, these changes seem pretty solid to me. I think it probably does make the most sense to have redcap_connect.php be its own file this way rather than in the EM class.

PR #238, opened 2019-12-20
Re-enable arrays and object as interpolation sources

grezniczek:
Hey @mmcev106 
thanks for the fix of the regexp string. I didn't think about escaping when changing from the literal to the string.

Your other changes, however, prevented the use of arrays and objects as sources for interpolation values. I have reintroduced `_getValues()` to handle this, but only in one place. The JS-`tt()` defined in _AbstractExternalModule_ stays as you proposed (I only changed the `var`s to `const`s to be more explicit).

While testing, I noticed that there was another bug in `interpolate` that would prevent escaping of curly braces when there were no interpolation values. I have changed this, so that the interpolation and excaping logic _always_ runs (provided string is indeed a string).

### DIFF
--- a/classes/AbstractExternalModule.php
+++ b/classes/AbstractExternalModule.php
@@ -1294,9 +1294,9 @@ class AbstractExternalModule
 				 * @returns {string} The interpolated string.
 				 */
 				module.tt = function (key) {
-					const argArray = Array.prototype.slice.call(arguments)
+					var argArray = Array.prototype.slice.call(arguments)
 					argArray[0] = this._constructLanguageKey(key)
-					const lang = window.ExternalModules.$lang
+					var lang = window.ExternalModules.$lang
 					return lang.tt.apply(lang, argArray)
 				}
 				/**
--- a/classes/AbstractExternalModule.php
+++ b/classes/AbstractExternalModule.php
@@ -1294,10 +1294,9 @@ class AbstractExternalModule
 				 * @returns {string} The interpolated string.
 				 */
 				module.tt = function (key) {
-					var argArray = Array.prototype.slice.call(arguments)
+					const argArray = Array.prototype.slice.call(arguments)
 					argArray[0] = this._constructLanguageKey(key)
-					
-					var lang = window.ExternalModules.$lang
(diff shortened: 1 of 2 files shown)


classes/AbstractExternalModule.php
```diff
@@ -1294,10 +1294,9 @@ public function initializeJavascriptModuleObject()
 				 * @returns {string} The interpolated string.
 				 */
 				module.tt = function (key) {
-					var argArray = Array.prototype.slice.call(arguments)
+					const argArray = Array.prototype.slice.call(arguments)
 					argArray[0] = this._constructLanguageKey(key)
-					
-					var lang = window.ExternalModules.$lang
+					const lang = window.ExternalModules.$lang
```
mmcev106: We still support back to IE 9 currently, so I'm afraid we can't use `const` :(

Would you prefer to change these back?  I'm also happy to do it after merging this.

classes/ExternalModules.php
```diff
@@ -576,15 +576,33 @@ public static function tt_initializeJSLanguageStore() {
...
 					 * Get and interpolate a translation.
 					 * @param {string} key The key for the string.
 					 * Note: Any further arguments after key will be used for interpolation. If the first such argument is an array, it will be used as the interpolation source.
 					 * @returns {string} The interpolated string.
 					 */
 					lang.tt = function(key) {
-						var string = this.get(key)
-						var values = Array.prototype.slice.call(arguments, 1)
+						const string = this.get(key)
+						const values = this._getValues(arguments)
```
mmcev106: See https://github.com/vanderbilt/redcap-external-modules/pull/238/files#r360443793

grezniczek:
I changed the `const`s to `var`s. Including the one that had slipped through before ,)  
https://github.com/vanderbilt/redcap-external-modules/pull/238/commits/9ec8f49c06596ea3f104d1616d7adffaa656d1a6

mmcev106:
@grezniczek, thanks very much for the quick response on this.  I went ahead and cherry-picked our three commits into the release branch so they'll make it into the upcoming LTS release.

PR #242, opened 2020-01-22
Prevented cron jobs from effecting each other by changing parameters

mmcev106:
I believe this should be safe across the board, but it definitely warrants review since the global `$_GET` and `$_POST` vars are being modified (only for crons).


taylorr4:
@mmcev106 these changes look fine to me. They don't worry me.

mmcev106:
Kyle just approved via Slack.  Merging...

PR #248, opened 2020-02-04
Ensure localization is initialized before used

grezniczek:
@mmcev106, while enhancing the Localization Demo EM for my upcoming talk this Friday, I noticed that we missed scenario, where translatable strings in `config.json` are used before the respective module is instantiated, and therefore the language strings are missing (and resulting in the _"Language key 'abc' is not defined"_ message). For example, this affected the name of links.

This pull request addresses the issue, by ensuring that initialization (loading of language strings) has indeed happened before config translation. To avoid double work, `initializeLocalizationSupport()` keeps track of the modules already initialized.

It would be nice if this PR could be fast-tracked into the next REDCap release. Thanks.

### DIFF
--- a/classes/ExternalModules.php
+++ b/classes/ExternalModules.php
@@ -165,7 +165,8 @@ class ExternalModules
 
 	private static $configs = array();
 
-	
+	// Holds module prefixes for which language strings have already been added to $lang.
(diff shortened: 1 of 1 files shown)


mmcev106:
Thanks @grezniczek!

Issue #261, opened 2020-02-14
Return value of new email hook is ignored, resulting in true being returned to send() in Message


grezniczek:
To quote from ExternalModules (about line 2647):
> // The following check assumes hook return values will always be arrays.
> // That's totally fine for now since our only hook that returns a value does return an array.
> // There may come a day when we'd want to support other types as well.

That day had arrived with the `redcap_email` hook :wink: 

As it does not return an array, this needs to be fixed. 
My proposal: limit the return value to strictly `true` or `false` and AND the results of potentially multiple calls of this hook in different modules together.

A fix for this is **urgent**, as now, the `recap_email` hook **cannot in any sensible way be used in the context of an EM.**
I have submitted a PR with a fix as described above. 

There still is the issue that multiple (unwanted) mails may be sent by modules that may send via some way or other, so maybe this hook signature should be extended by a "handled" parameter that indicates to the next module in the chain whether the email has already been handled. Alternatively, hook processing could stop after the first `false`. However, emails may then never get to the (by the user, admin, ..) intended module. This definitely needs some more consideration. 

_Maybe something like this works:_ Hook processing stops at the first `false`, and modules need to  announce (via config.json) whether they act as email exit gateways (i.e. the email leave the system via a mechanism implemented by the module) or simply as email modifiers (i.e. they modify any of the email items and reinject the email back via a call to `REDCap::email()` and returning `false`). The modifiers should then be called first, and maybe admins need a way to determine the order.

123andy:
just glancing on my mobile but the array return values are for the custom user verify hook.  maybe a similar json return is better with a result key : success or error,  and then a message key.    seems more extensible.

________________________________

grezniczek:
As the return value ultimately is interpreted by REDCap core in Message->send(), the value delivered back by the EM framework needs to be either `true` or `false`. 

To elaborate more on the issue - what really is missing is a mechanism to handle situations like this:

-  One module's job is to add some footer material (set in config) at the bottom of every email sent from the project (Modifier-EM). 
- Another module's job is to send emails through different SMTP hosts, depending on sender address, just like in the examples for this hook (Sender-EM).
- A third module's job is to simply log each email  being sent by calling into some external service (Logger-EM)

How can this be achieved without duplicate emails being sent and all modules getting a chance to do what they are supposed to?

[...]

PR #262, opened 2020-02-14
Fix email hook

grezniczek:
Fix for #261

### DIFF
--- a/classes/ExternalModules.php
+++ b/classes/ExternalModules.php
@@ -2653,10 +2653,13 @@ class ExternalModules
 						"result" => $result
 					);
 				}
+				return $result;
 			};
 
 			foreach($versionsByPrefix as $prefix=>$version){
-				$startHook($prefix, $version);
+				$result = $startHook($prefix, $version);
+				if ($name == "email" && $result === false) 
+					break;
 			}
 
 			$callDelayedHooks = function($lastRun) use ($startHook){
--- a/classes/ExternalModules.php
+++ b/classes/ExternalModules.php
@@ -2653,13 +2653,10 @@ class ExternalModules
 						"result" => $result
 					);
 				}
-				return $result;
 			};
 
 			foreach($versionsByPrefix as $prefix=>$version){
-				$result = $startHook($prefix, $version);
-				if ($name == "email" && $result === false) 
-					break;
+				$startHook($prefix, $version);
 			}
 
 			$callDelayedHooks = function($lastRun) use ($startHook){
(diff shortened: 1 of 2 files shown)


grezniczek:
Commit https://github.com/vanderbilt/redcap-external-modules/pull/262/commits/947844c8783244182f723222dcc6db1a4f2d9fe7 adds that the email processing stops after the first `false`. Consider this _optional_, but it may be safest for now to prevent potentially unwanted sending of duplicate emails.

classes/ExternalModules.php
```diff
@@ -2644,20 +2644,22 @@ static function callHook($name, $arguments, $prefix = null)
...
 					);
 				}
+				return $result;
 			};
 
 			foreach($versionsByPrefix as $prefix=>$version){
-				$startHook($prefix, $version);
+				$result = $startHook($prefix, $version);
+				if ($name == "email" && $result === false) 
+					break;
```
mmcev106: I don't think we want to break here.  We could potentially have a module that returns false, then another modules that runs after it, gleans some information from the email being sent, then returns true.

grezniczek: Well, in this case, maybe just ignore the last two commits https://github.com/vanderbilt/redcap-external-modules/pull/262/commits/947844c8783244182f723222dcc6db1a4f2d9fe7 and https://github.com/vanderbilt/redcap-external-modules/pull/262/commits/e6dc40c64c197c0339a74261c96312f7d19eabe0

In this case it may be a good idea, however, to add another parameter to the email hook indicating to subsequent modules that some module prior has already "taken care of sending" (by returning `false`), something like `$handled` which will be `false` by default and set to `true` by the processing loop when a module returns `false`?

grezniczek: As the signature of the hook function should probably be consistent between the framework and REDCap hooks, this would have to be discussed with @taylorr4

grezniczek:
I reverted the commit that added the "exit after first false" behavior (and the corresponding change to the docs).

classes/ExternalModules.php
mmcev106: Resolving... @grezniczek, reverted those last two commits.  We may consider alternatives in the future.

mmcev106:
Merged!  Thanks you @grezniczek.  This is very much appreciated.  If possible it'd be great if this could go through our normal release process and sit on our testing server for a week.  However, do you or @taylorr4 think we should go ahead and merge this to the release branch?

grezniczek:
I guess there are no (publicly released) modules yet that make use of this hook, so I guess it's fine to let this go the usual way.

More consideration regarding this particular hook is needed anyways. See https://github.com/vanderbilt/redcap-external-modules/issues/261#issuecomment-586716341.

PR #263, opened 2020-02-17
Removed a new features block from v3 since the features are now available in v2 as well

mmcev106:
All new framework methods are now available when using v2 (on whatever REDCap version added each feature).  It seems like this info is already covered elsewhere in the docs (like other methods).  However, I figured I'd run this by you in case you thought we tweak the other docs at all.


grezniczek:
Looks great. I added a few more updates following along: #264.

mmcev106:
Thanks much!

PR #268, opened 2020-03-01
Limit config translation

grezniczek:
Translation of `config.json` via the _tt__ prefix would allow to translate anything, including items that should not be translated (as author details) and items that must not be translated (such as keys). 

In order to avoid confusion and potential misuse, this PR limits the applicabilty of config translations by providing a whitelist of keys and a blacklist of sections.


mmcev106:
It looks like merging this broken the `testTranslateConfig()` unit test for translating project setting names.  Is that valid use case?

grezniczek:
I'll check.

grezniczek:
I found the bug ... see #272

mmcev106:
It also looks like our code scanner detected that array constants are not allowed in PHP 5.5...  Do you mind converting those to variables?  It might be time to time to ask Rob about bumping the REDCap minimum up to 5.6...

grezniczek:
Done. #273

PR #269, opened 2020-03-01
Minor doc updates

grezniczek:
I used the `User` and `Project` objects for the first time and noticed that their fields are not documented. I added this in this PR as well as a few other minor enhancements (mainly for consistent looks).

@mmcev106, did you mean the `framework` reference in these objects to be private? Maybe the classes should then be amended to specify the privacy of this field?

### DIFF
--- a/classes/framework/Project.php
+++ b/classes/framework/Project.php
@@ -23,4 +23,8 @@ class Project
 
 		return $users;
 	}
+
+	function getProjectId() {
+		return $this->project_id;
+	}
 }
--- a/classes/framework/User.php
+++ b/classes/framework/User.php
@@ -39,6 +39,10 @@ class User
 		return $this->user_info;
 	}
 
+	function getUsername() {
+		return $this->username;
+	}
+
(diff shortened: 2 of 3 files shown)


mmcev106:
Any undocumented variables/methods are intentional, and are not currently supported (they may be refactored at any time).

grezniczek:
Wouldn't _username_ and _project_id_, respectively, be sensible documented fields in these objects?

If I am not mistaken, there currently is no method (like getProjectId) to get the name of the current user (except from the USERID constant). IMO it would make sense to have this exposed on a `User` object.

Anyways, I can eliminate the fields, if you want, so that at least the cosmetic (and version info) additions can me merged.

mmcev106:
I think it would be more appropriate to create getter methods for each, as opposed to exposing them directly.  That's our standard or other "static" values in other documented methods, and provides more flexibility for future refactoring.

grezniczek:
How about this?

mmcev106:
Looks great.  Thanks!

PR #271, opened 2020-03-02
Change ? to ! for future REDCap releases

grezniczek:
I was just thinking .... to make release version tracking easier ... why not mark **_future_** REDCap releases with bangs instead of question marks?

This way, one "only" needs to do a search for `!.!.!` in the framework docs after a REDCap release and replace it with the current version.

Or even **better**, this could be integrated into the REDCap build script (I am assuming something like this exists) and be done automatically. `!.!.!` in markdown files inside `ExternalModules` should be pretty unique (I tested and found only the two occurences introduced with this PR).


mmcev106:
I'm trying to move us toward an automated solution for filling out these versions (slowly but surely).  I think the bang vs. question mark idea is interesting, but it might actually complicate things for a scripting process.  The minute we have such a process in place, I'm going to either automatically or manually go through all old methods and fill out their versions (so there's be no need to distinguish between the two).  I'm going to close this, but feel free to reopen it now or in the future for further discussion.

PR #272, opened 2020-03-02
Fix translateConfig

grezniczek:
I am not sure I totally understand why this is needed, but the fact is that it is: using strict comparison for the two `in_array` calls.
This fixes the broken unit test mentioned here: https://github.com/vanderbilt/redcap-external-modules/pull/268#issuecomment-593571292


grezniczek:
Ah well, I do understand after all ... `0 == "something that certainly does not look like zero"` evaluates to `true`. I guess it's time I get the hang on this.

mmcev106:
I forgot that third argument to `in_array()` existed.  Good to know...

PR #275, opened 2020-03-03
Repeating branching logic

ihabzee:
Hi @mmcev106 

I created new PR just for the branching logic. hope this is better for you. Also I ignored `.editorconfig` for now. Maybe it would be good practice to prevent such an issue from showing up again :) 

Please let me know if you have any question. 
Thanks,


mmcev106:
@ihabzee, are you aware that this new PR still changes over 1000 lines (see the "Files Changed" tab?  Would it be possible to disable your code formatter temporarily, or set it to only format lines you edit (many formatters have that feature these days).  I do appreciate the pull requests, but I'm not sure this level of code format changes are appropriate without code standard discussions.

mmcev106:
I'm curious, if you set all your `.editorconfig` values to `none`, will it effectively disable the formatter (only for the EM framework)?  Here's an example for "ignoring files":

https://stackoverflow.com/a/40348831/2044597

ihabzee:
even if I set all attributes in `.editorconfig` to none the problem will persist since my IDE reformat on commit. I have to apply my changes manually and disable format on commit.

ihabzee:
@mmcev106  I tried to clean it as much as I can would you mind looking at it now? 

Thanks

mmcev106:
Thanks very much @ihabzee!  The diff is much more digestible now.

I'm working on creating a basic unit test for setting branching logic.  Once finished, I'll run that tests on this branch, and go ahead and merge it if they pass.

manager/js/globals.js
```diff
@@ -431,11 +533,11 @@ ExternalModules.Settings.prototype.getColumnHtml = function(setting,value,classN
...
 	}
 	var trClass = className;
-	
+
 	var colspan = '';
-	if(type == 'descriptive'){
+	/*if(type == 'descriptive'){
 		colspan = " colspan='3'";
-	}
+	}*/
```
mmcev106: It looks like this might break descriptive fields.  I just added a descriptive field to the example module in the testing branch if you'd like to use that to test.

manager/js/globals.js
```diff
@@ -540,15 +642,17 @@ ExternalModules.Settings.prototype.getColumnHtml = function(setting,value,classN
...
-	
-	if(type != 'descriptive'){
+
+    html += "<td class='external-modules-input-td'>" + inputHtml + "</td>";
+
+	/*if(type != 'descriptive'){
 		html += "<td class='external-modules-input-td'>" + inputHtml + "</td>";
-	}
-	
+	}*/
```
mmcev106: It looks like this might break descriptive fields.  I just added a descriptive field to the example module in the testing branch if you'd like to use that to test.

manager/js/globals.js
```diff
@@ -883,6 +987,39 @@ ExternalModules.Settings.prototype.resetConfigInstances = function() {
...
+
+};
+
+/**
+ * search object for a key
+ * @param obj
+ * @param query
+ * @returns {*}
+ */
+ExternalModules.Settings.prototype.searchConfigViaKeys = function (obj, query) {
```
mmcev106: It looks like this method is not used anywhere.  Should it be removed?

ihabzee:
@mmcev106 try below a config.json example of use case that wont work before PR but will work after. 

`	"project-settings": [
      {
        "key": "top-level",
        "name": "Top Level",
        "type": "checkbox"
      },
      {
        "key": "countries",
        "name": "Country",
        "type": "sub_settings",
        "repeatable": true,
        "sub_settings": [
          {
            "key": "country-name",
            "name": "Country Name (on the bottom just to test a different order)",
            "type": "text"
          },
          {
            "key": "states",
            "name": "State",
            "type": "sub_settings",
            "repeatable": true,
            "branchingLogic": {
              "field": "country-name",
              "op": "=",
              "value": "USA"
            },
            "sub_settings": [
              {
                "key": "state-name",
                "name": "State Name",
                "type": "dropdown",
                "choices": [
                  {
                    "name": "Arizona",
                    "value": "AZ"
                  },
                  {
                    "name": "California",
                    "value": "CA"
                  }
                ]
              },
              {
                "key": "state-top-level",
                "name": "Top Level is True",
                "type": "descriptive",
                "branchingLogic": {
                  "field": "top-level",
                  "op": "=",
                  "value": "1"
                }
              },
              {
                "key": "cities",
                "name": "City",
                "type": "sub_settings",
                "repeatable": true,
                "branchingLogic": {
                  "field": "state-name",
                  "op": "!=",
                  "value": ""
                },
                "sub_settings": [
                  {
                    "key": "city-name",
                    "name": "City Name",
                    "type": "text"
                  },
                  {
                    "key": "desc-california",
                    "name": "You are in California",
                    "type": "text",
                    "branchingLogic": {
                      "field": "state-name",
                      "op": "=",
                      "value": "CA"
                    }
                  },
                  {
                    "key": "desc-usa",
                    "name": "You are in the USA",
                    "type": "descriptive"
                  }
                ]
              }
            ]
          }
        ]
      }
	],`

manager/js/globals.js
mmcev106: I've confirmed that descriptive fields are now displayed again.

manager/js/globals.js
mmcev106: This method was indeed removed.

manager/js/globals.js
mmcev106: I am unresolving this because it was not addressed.  This changes causes descriptive fields to no longer take up the full width of the configuration dialog.  I see where you added a fake input field which I'm guessing helps branching logic apply to descriptive fields.  Is that the only solution?  If so, would it make sense to put the input inside the first `<td>` alongside the descriptive text, and reintroduce the `colspan`?

manager/js/globals.js
```diff
@@ -523,7 +625,10 @@ ExternalModules.Settings.prototype.getColumnHtml = function(setting,value,classN
 	}
 	else if(type == 'custom') {
 		inputHtml = this.getInputElement(type, key, value, inputAttributes);
-	} else {
+	}else if (type == 'descriptive') {
+        //add hidden field for descriptive type to trigger correct branching logic when config page is loaded.
+        inputHtml = this.getInputElement('hidden', key, value, inputAttributes);
```
mmcev106: This line causes the empty string for this field to be sent back to the server for descriptive fields when settings are saved.    Is it possible to produce the desired behavior without adding an input for descriptive fields?

mmcev106:
I'm sorry for the delays in reviewing this.  I'm juggling many different priorities...  Thanks very much for the configuration example.  I'll try to incorporate it into the unit test.

mmcev106:
While testing, I'm wondering if doBranching() should fire as each character is typed on text fields, instead of after the field loses focus.  @ihabzee, what do you think about that?

mmcev106:
I'm sorry, I think I also just found a bug...  In the provided config.json example changing `top-level` causes `state-top-level` to appear even when it's parent `states` is still hidden.

mmcev106:
I wonder if automatically ANDing the logic of any subsettings with their parents recursively would be an easy way to fix that...

manager/js/globals.js
ihabzee: We need this hidden field so can calculate the level of the descriptive field and show/hide based on branching logic.

manager/js/globals.js
ihabzee: I reverted this.

ihabzee:
@mmcev106 I added a fix for grandchildren logic. can you test one more time? Thanks for your help.

manager/js/globals.js
mmcev106: So, simply reverting this means the new input for descriptive fields adds some margin to the right side of the entire config dialog.  Would a simple solution be simply placing the input inside the first <td> for descriptive fields (instead of inside it's own <td> like other fields)?

manager/js/globals.js
mmcev106: Would it be possible for branching logic to trigger of a class (perhaps `external-modules-input-element`) instead of a particular element type (like `<input>`)?  If so, then this new hidden field could simply be a `<div>`, which would not get included when sending setting values back to the server.

mmcev106: Perhaps this is already partially implemented since we support other element types like `<select>` or `<textarea>` already.

mmcev106:
I'm sorry, I'm afraid I'm still seeing the same problem: https://github.com/vanderbilt/redcap-external-modules/pull/275#issuecomment-599661346

ihabzee:
@mmcev106  Sorry for the delay. I pushed another fix. If this does not work please give me the steps to re-produce. 
Thanks,

mmcev106:
It unfortunately looks like a recent commit breaks basic top level branching logic.  To reproduce this:
1. Merge the latest version of the `testing` branch
1. Enable the bundled "Configuration Example" module and go to it's system settings
1. Enter "1" for the value of the "System Text" setting.  You'll notice a javascript error when attempting to display the "branching-logic-example-1" setting.

You may want to start running the JS unit tests before committing to get immediate feedback instead of waiting on me to run them.  A "Click here to run JS Unit Tests" link should be visible on the Control Center External Modules page if the `isLocalHost()` check is returning true.  You could hack that method to return true if you're running on a different domain (or include support for that domain in this PR).  Interpreting JS unit test failures involves stepping through stack traces currently (there's a learning curve).  Let me know if you have any questions there.

ihabzee:
> It unfortunately looks like a recent commit breaks basic top level branching logic. To reproduce this:
> 
> 1. Merge the latest version of the `testing` branch
> 2. Enable the bundled "Configuration Example" module and go to it's system settings
> 3. Enter "1" for the value of the "System Text" setting.  You'll notice a javascript error when attempting to display the "branching-logic-example-1" setting.
> 
> You may want to start running the JS unit tests before committing to get immediate feedback instead of waiting on me to run them. A "Click here to run JS Unit Tests" link should be visible on the Control Center External Modules page if the `isLocalHost()` check is returning true. You could hack that method to return true if you're running on a different domain (or include support for that domain in this PR). Interpreting JS unit test failures involves stepping through stack traces currently (there's a learning curve). Let me know if you have any questions there.

I was able to run the unit test and fix the problem occurred there. but the unit test stills failed on theme.js and tinymce.js 

Also I tested the EM you mentioned and when I add 1 to System Text I got `Visible When System Text is '1':` I think this is intended behavior correct? 

If you would like we can jump on few minute call to go over any issue you still facing. 

Thanks,

mmcev106:
I'm sorry for the delay @ihabzee.  I need to focus on a COVID related and am not sure when I'll get back to this just yet.

mmcev106:
I'm going to close this as @grezniczek's [new settings configuration dialog](https://github.com/grezniczek/redcap_em_config_study) will likely be completed in the coming months and resolve many current branching logic issues.  Don't hesitate to reopen and/or ask questions if anyone has additional thoughts/concerns.

Issue #276, opened 2020-03-04
setProjectSettings() broken


grezniczek:
It seems the (documented, framework **v1**) function `setProjectSettings()` is broken (lines 217ff in AbstractExternalModule.php; both _testing_ and in REDCap 9.7.7):

```php
/**
 * Saves all project settings (to be used with getProjectSettings).  Useful
 * for cases when you may create a custom config page or need to overwrite all
 * project settings for an external module.
 * @param array $settings Array of all project-specific settings
 * @param int|null $pid
 */
function setProjectSettings($settings, $pid = null)
{
	$pid = self::requireProjectId($pid);
	ExternalModules::saveSettings($this->PREFIX, $pid, json_encode($settings));
}
```

The **json_encode** is quite out of place, as `saveSettings() `expects an array as third argument. Thus, this leads to nothing being processed at all.

Which turns out to be a good thing, as the downstream processing of an actual array (like the one from `getProjectSettings()`, after removing json_encode in the snippet above) screws up the data stored in the _redcap_external_module_settings_ ... **badly**.

mmcev106:
None of the modules on our instance reference this method.  I wonder if the best course of action would be to remove it from the docs, and make it email the system admin with a deprecation warning just in case anyone is actually using it.  If we don't hear of any reports of the email firing, we could remove the method entirely at some point in the future.

123andy:
I have vague memories of trying to use this and punting.  I do think it would be a valuable method for 'custom' configuration pages.  If we could make a simple single call to take the config params from an ajax post and dump them to the EM settings that would be convenient...

grezniczek:
I guess nobody is actually using it as it simply does nothing.

But I just had another look at it. It would seem the issue with corruption I saw after removing json_encode and passing the data obtained with `getProjectSettings()` to `saveProjectSettings()` is caused by two things: 
1)  `getProjectSettings()` is calling `ExternalModules::getProjectSettingsAsArray()` without specifying `false` as the third argument, thus returning project **AND** system settings. 
2) The raw settings that are expected by `ExternalModules::saveSettings()` are structured totally different from what getProjectSettings() returns, or even what `ExternalModules::formatRawSettings()` transforms them into.

It would seem the solution is to transform the result from `getProjectSettings()`  (which we probably cannot change now) and which, as per docs, is supposed to be the input (after modification) to `setProjectsSettings()`, into something that matches the output of `ExternalModules::formatRawSettings()`.

[...]

PR #277, opened 2020-03-05
Fix get/setProjectSettings()

grezniczek:
See discussion here: #276 

This adds working versions to the v4 framework, leaving v1-3 behavior as is.
(I hope I got the framework versioning right, as a lot has changed here recently).


grezniczek:
@mmcev106 any chance of getting this into 9.7.8 as well? If not, this may need to be changed to v5.

mmcev106:
I'm afraid not.  The changes going into 9.7.8 have been in testing for a few weeks.  Making v5 for this makes sense to me.

grezniczek:
This is now updated to v5 and current with testing.

docs/framework/v5.md
```diff
@@ -0,0 +1,10 @@
+## Framework Version 5
+
+See the [Framework Intro](intro.md) page for more details on framework versions in general.
+
+Framework Version 5 was first added to REDCap in Standard Release ?.?.?. It first appeared in REDCap LTS at ?.?.?.
```
mmcev106: I actually just removed these lines in each version's md file, since they duplicated the info on the intro page linked above.  I'll go ahead and merge this and remove it myself though.

PR #282, opened 2020-03-13
Fixed project setting typeaheads

mmcev106:
Apparently I broke typeaheads for project id settings back in January.  Simplifying the the project id setting rendering code to make it work like other typeaheads was the simplest way to fix it.

@kcmcg, if you're swamped, is there anyone else on our team who could review this?  We could ask around in our Monday meeting if anyone has unbillable time available.


mmcev106:
The list is limited to projects for which the user has design rights.  For superusers, it would be the same list they's see on their "My Projects" page.  I just spoofed about 1300 options in the dropdown.  On my 8 year old laptop that slowed the initial dropdown display down to about one second.  Typing is a little sluggish for the first couple of characters.  I think I'd consider that level of slowdown acceptable.  What do you think?

kcmcg:
Agreed. That seems acceptable to me.

PR #283, opened 2020-03-18
Replaced the shutdown check with a wider reaching recursive hook check

mmcev106:
@scottjpearson, would it be worth your time to review this perchance?  I'm trying to get in the habit of getting at least one other set of eyes on higher risk changes.  You probably know the most about module hook calls of everyone else on the team.

### DIFF
--- a/classes/ExternalModules.php
+++ b/classes/ExternalModules.php
@@ -172,7 +172,6 @@ class ExternalModules
 
 	private static $initialized = false;
 	private static $activeModulePrefix;
-	private static $shuttingDown = false;
 	private static $instanceCache = array();
 	private static $idsByPrefix;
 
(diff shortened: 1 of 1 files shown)


grezniczek:
Does this work "per module" or "per (original) email being sent"? Put otherwise, if both, module A and B subscribe to the email-hook (assuming A being processed before B), will B see an email modified and re-sent via REDCap::email by A? The modified version? And not only the original version?

This ties in with #261.

I could try to find a bit of time to work on a PR regarding this. Do you think this would be feasible as outlined in #261?

mmcev106:
@grezniczek, I'm afraid I don't currently understand all possible cases well enough to answer your questions definitively.  It depends on what each module's `redcap_email` hook does...  I do feel confident enough to go ahead and merge this to allow `redcap_email` hooks in modules to trigger errors and call `REDCap::email()` without causing an infinite loop.

PR #284, opened 2020-03-21
Fix for link icons

123andy:
With the version3+ version you could no longer use built-in redcap .png files for link icons.  This allows the use of all icons.  In fact, it could probably be retrospectively added to the original and not just be a version 3 modification.

### DIFF
--- a/classes/ExternalModules.php
+++ b/classes/ExternalModules.php
@@ -5062,14 +5062,33 @@ class ExternalModules
 		};
 
 		if(ExternalModules::getFrameworkVersion($module) >= 3){
-			$iconPath = $module->framework->getModulePath() . '/' . $icon;
-			if(file_exists($iconPath)){
-				$iconElement = $getImageIconElement($module->getUrl($icon));
-			}
-			else{
-				// Assume it is a font awesome class.
+			do {
+				// 1: Try inside the module first
+				$iconPath = $module->framework->getModulePath() . '/' . $icon;
+				if(file_exists($iconPath)) {
+				    $iconElement = $getImageIconElement($module->getUrl($icon));
+				    break;
+				}
+
+				// 2: Try the EM built-in icons
+				$iconPathSuffix = 'images/' . $icon . '.png';
+				if(file_exists(ExternalModules::$BASE_PATH . $iconPathSuffix )) {
+				    $iconUrl     = ExternalModules::$BASE_URL . $iconPathSuffix;
+				    $iconElement = $getImageIconElement($iconUrl);
+				    break;
+				}
+
+				// 3: Try the built-in REDCap icons
+				global $redcap_version;
+				if(file_exists(APP_PATH_DOCROOT.DS."/Resources/".$iconPathSuffix)) {
+				    $iconUrl     = APP_PATH_WEBROOT . 'Resources/' . $iconPathSuffix;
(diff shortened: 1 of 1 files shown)


mmcev106:
@taylorr4, what are your thoughts on this change?  If I remember correctly, I think the main reason we introduced the icon change as a breaking change in framework version 3 was to force module authors to use font awesome instead of the old bundled icons (but still allow them to supply their own).  Are those old icons going to disappear at some point?  Was it just for consistency's sake, or was framework version 3 a bit heavy handed in hindsight?

taylorr4:
@mmcev106 I think these changes are fine. But honestly, as we use font awesome more and more, we have actually removed some old png icons over time. We're not actively removing them, but they could disappear at any moment (not all but a few from time to time). So it's probably fine to allow module authors to still reference the old png icons in REDCap if they want, with the caveat that they might not always be there. But that seems like it has minimal consequences since the author can easily push out a new module version to fix the icon in the unlikely event that that ever happens. So i'm not too worried about allowing authors to reference REDCap icons.

mmcev106:
Hmm, that makes me think we should leave the current behavior in place.  As a module author, I would be annoyed if the icon I was using suddenly disappeared one day.  I'd much prefer to choose whether to copy the icon into my module, or switch to a font awesome icon at a time of my choosing (when updating to framework version 3).

mmcev106:
@123andy, what are your thoughts after considering the goals/risks that lead to this change initially?

123andy:
@mmcev106 - In my case, I wanted a big red icon to grab attention and didn't find it immediately in font-awesome so I just kept what I had used before and was surprised when it didn't work.  I assumed it was oversight that the change broke previous behavior as it wasn't documented at all.  

If REDCap drops a common .png image, it is going to break all the pre-version 3 EMs that use them already.  If we really are moving on a path of deprecating resource images, then I'm fine with it.  If we are not, then I'd say the breaking change was unnecessary.  I don't feel that strongly about it.  I has assumed the break was just an oversight and intentional when I came across it.

mmcev106:
@123andy, I thought this breaking change was pretty well documented in the breaking change list for the [v3 docs](https://github.com/vanderbilt/redcap-external-modules/blob/testing/docs/framework/v3.md)...  Do you have any recommendations on how to clarify this, or what to do when updating framework versions in general?  I fear I spend too much time in the trenches to have a good perspective on how others view such things.

123andy:
As usual, you're right.  I only read this part:

https://github.com/vanderbilt/redcap-external-modules/blob/testing/docs/official-documentation.md#2-configuration-file

which doesn't make it sound like a 'breaking change'.  I guess I could make a pull request to clarify.

mmcev106:
[Click here](https://github.com/vanderbilt/redcap-external-modules/commit/a19a2598e4561e2959fcb489c2b2ae600c2a51c6) for my attempt to clarify the docs.  I have mixed feelings about merging the `official-documentation.md` and `framework/intro.md`.  The former seems more intended for new developers or deep dives, while the latter is like a quick reference.  I wonder if it make sense to move `intro.md` up a directory and call it `methods.md` now that the method docs have been moved there.

I'll go ahead and close this for now, but please feel free to speak up if anyone has any more ideas to help clarify things.

PR #290, opened 2020-03-31
Public survey url by project

knil-maloon:
This would allow to get the link for any project given the project_id and instrument name and I also added the option to get only the code.


classes/AbstractExternalModule.php
```diff
@@ -1112,9 +1112,14 @@ public function redcap_module_configure_button_display(){
         return true;
     }
 
-    public function getPublicSurveyUrl(){
-        $instrumentNames = \REDCap::getInstrumentNames();
-        $formName = key($instrumentNames);
+    public function getPublicSurveyUrl($pid=null,$instrument=null,$hash=false){
+        if(empty(!$pid) && !empty($instrument)){
```
mmcev106: I'm a little confused about the instrument parameter.  Can't there only be one instrument with public survey url?  If it is possible for multiple instruments to have public survey links, then should we make this method default to the first one, especially if it's the only one?

classes/AbstractExternalModule.php
```diff
@@ -1112,9 +1112,14 @@ public function redcap_module_configure_button_display(){
         return true;
     }
 
-    public function getPublicSurveyUrl(){
-        $instrumentNames = \REDCap::getInstrumentNames();
-        $formName = key($instrumentNames);
+    public function getPublicSurveyUrl($pid=null,$instrument=null,$hash=false){
```
mmcev106: This is a reminder for myself to run unit tests prior to merging this PR.

classes/AbstractExternalModule.php
mmcev106: Should the first `!` sign be on the `empty()` call instead of the `$pid` param itself?

docs/framework/intro.md
```diff
@@ -60,7 +60,7 @@ getProjectId() | `?.?.?`<br>`1` | A convenience method for returning the current
...
 getProjectSetting($key&nbsp;[,&nbsp;$pid]) | `?.?.?`<br>`1` | Returns the value stored for the specified key for the current project if it exists.  For non-repeatable settings, `null` is returned if no value is set.  For repeatable settings, an array with a single `null` value is returned if no value is set.  In most cases the project id can be detected automatically, but it can optionally be specified as a parameter instead.
 getProjectSettings([$pid]) | `?.?.?`<br>`1` | Gets all project settings as an array.  Useful for cases when you may be creating a custom config page for the external module in a project. **Breaking change in framework v5.**
-getPublicSurveyUrl() | `?.?.?`<br>`1` | Returns the public survey url for the current project.
+getPublicSurveyUrl($pid=null,$instrument=null,$hash=false) | `?.?.?`<br>`1` | Returns the public survey url for the current project. If $hash is true it will return only the hashcode. By default it will be false returning the link. If a project_id and an instrument have a value it will return the link for that specific project.
```
mmcev106: I'm not sure having a `$hash` parameter change the `getPublicSurveyUrl()` method so it no longer returns a URL is quite appropriate.  I think it might be preferred to have a separate `getPublicSurveyHash()` method that gets the hash.  The URL method can simply call this method and prepend the URL before returned (should just be a line or two).

classes/AbstractExternalModule.php
knil-maloon: As it's another project we probably can't use the getInstrumentNames as it would return from the current project and not another so we need both parameters

knil-maloon: Yes, that was a typo on my end...

docs/framework/intro.md
knil-maloon: That's a good idea :)
I also thought it wasn't that good with the function having that name but I liked the option.

classes/AbstractExternalModule.php
mmcev106: I just made a commit changing `ExternalModules::getFormNames()` to accept a `$pid` argument.  If you merge the latest version of testing into this branch, can that function be used instead of `getInstrumentNames()`?

knil-maloon: I'll test it!

knil-maloon: Is it already on test? My code didn't update that function

mmcev106: I pushed it to the testing branch before mentioning it here.

knil-maloon: with this change we need to change the method call to get the first survey element as we can't do an sql with an array of form names...

mmcev106: After thinking about this again, only one form can have a public survey link, so can't we just return the form name from the existing query instead of calling any other method to get form names?

knil-maloon: you mean making a sql first for the pid to return the form and then return that and use it as a name? Could it be that you can have more surveys for one pid?

grezniczek: If I may chip in ... apparently public surveys are identified in the backend database by having _participant_email_ being `null`. If this is correct, wouldn't a simplified JOIN with only redcap_surveys and redcap_surveys_participants tables (on survey_id) and a WHERE with project_id and "_participant_email_ is null" be sufficient to get the hash? 
The instrument name should be irrelevant, as there can ever be at most one public survey per project.

mmcev106: I believe the query in the code is already doing that.  That's not really what this review is about.

mmcev106: @knil-maloon, yes, doing the query first and returning the form_name it uses should work.  Projects can only have one form with a public survey link.

grezniczek: @mmcev106 The query in the code unnecessarily involves the redcap_metadata table. As you pointed out yourself in https://github.com/vanderbilt/redcap-external-modules/pull/290#discussion_r401230875, the instrument name is irrelevant. So why would it need to be retrieved?

mmcev106: Yes, you're spot on @grezniczek.

knil-maloon: Changed and added the new hash function

grezniczek: @knil-maloon, this is the reduced SQL:
```sql
select p.hash 
from redcap_surveys s
join redcap_surveys_participants p
on s.survey_id = p.survey_id
where s.project_id = ? and p.participant_email is null
```
You should call getPublicSurveyHash() from inside getPublicSurveyUrl() to avoid code duplication.

Furthermore, there is no handling of the case where zero rows are returned (i.e. there is no public survey for the project). I guess it would be ok to return `null` in this case for both functions (needs to be documented).

knil-maloon: I added to return null in both functions if the results are 0

mmcev106: Looks great!

classes/ExternalModules.php
```diff
@@ -5739,16 +5739,16 @@ public static function setRecordCompleteStatus($projectId, $recordId, $eventId,
...
 
-	public static function getFormNames(){
-		$metadata = self::getMetadata(TEST_SETTING_PID);
-		
-		$formNames = [];
-		foreach($metadata as $fieldName => $details){
-			$formNames[$details['form_name']] = true;
-		}
+    public static function getFormNames($pid){
+        $metadata = self::getMetadata($pid);
```
mmcev106, line 5743: I think this change might have been made accidentally.  It is causing a unit test to fail.

mmcev106: This looks like a partial change might have been manually copied over from the testing branch.  There is currently a conflict.  Merging the latest version of the testing branch and resolving that conflict is required before merging this anyway.  It will likely fix the unit test issue caused by this change.

knil-maloon: Can you check it now?

classes/AbstractExternalModule.php
grezniczek: Hmm, I just did a quick test ... when an instrument that was not the first (public) survey is "promoted" to be the first instrument (and thus becomes the new public survey), the `redcap_surveys_participants` table will contain more than just one row with _participant_email_ set to NULL. Thus, the first instrument becomes relevant, as there may be multiple entries that satisfy the shorter query.

This query should do the trick, still requiring only project_id as parameter:
```sql
select p.hash 
from redcap_surveys s
join redcap_surveys_participants p
on s.survey_id = p.survey_id
join redcap_metadata  m
on m.project_id = s.project_id and m.form_name = s.form_name
where p.participant_email is null and m.field_order = 1 and s.project_id = ?
```

knil-maloon: @grezniczek good catch. I thought you could only have one. I just changed the sql.

mmcev106: The `field_order = 1` check is what we've had in the testing branch for years, but I think that might not even be correct...  It looks like it does not cover the case where the first survey is on the second form (the first form does not have a survey).  I think we might have to `order by field_order` then `limit 1` to get the first.  May I have a second or third opinion on that?

grezniczek: @mmcev106, for a survey to be _public_, it **must** be the first form (and thus contain the record id field, which is **always** the first field. So I guess the query is fine, unless the previous statement is not true in all cases.

mmcev106: Spot on @grezniczek, thank you!  Unit tests are passing.  I'll merge this shortly.

classes/AbstractExternalModule.php
mmcev106: Done!

PR #293, opened 2020-04-01
Prevented the "->framework" syntax from being required for newer module methods

mmcev106:
I found an alternate solution for the main concern preventing us from doing this previously (see the isTesting() check & exception).  With that in place, I don't think there's any reason not to allow & the shorter syntax.

Long term, I'd like to move all methods from `AbstractExternalModule` to `Framework` to prevent modules from overriding them and interfering with core framework calls (like `getSystemSetting()`).  I moved `getRecordId()` and `setRecordId()` as an example to make sure that plan and the added exception were solid.

Any concerns?


mmcev106:
I just added a commit that removes framework versions from the method documentation.  After this and other recent changes, it is only required to know the REDCap version when each method was added (new methods are available on all framework versions).  I'll replace the "?.?.?" in the note about using `->framework` syntax once this change makes it into a release.

mmcev106:
Typo fixed.  Thank you for the review @grezniczek!

PR #299, opened 2020-04-09
New enable module function

knil-maloon:
To enable an EM within a different EM


knil-maloon:
I took out the "inside" word. I think now makes more sense

PR #300, opened 2020-04-10
Fixed a bug where Array function names could be interpretted as setting keys

mmcev106:
This was a weird one...  If y'all could help me verify that this has no potential to change any behavior outside of the specific case described below, I would very much appreciate it.

I ran into a case where the add instance button did nothing on a module with a setting named `keys`.  Prior to this change, we were getting `savedSettings['keys']` thinking it was returning a saved setting value, even though `savedSettings` was actually an Array (not a simple object) and we were getting a reference to the Array class's `keys()` function.  This would only have been a problem since last June when commit 4844098 unexpectedly bypassed a check that made this a non-issue previously.


123andy:
I've seen a few odd behaviors when using the word 'keys' or 'key' - thinking of sql around the external_module_log_parameters tables...  I've had to escape them as `key` in order to get SQL to work, but nothing like this.

PR #302, opened 2020-04-13
Enabled modules

grezniczek:
I needed to check whether another module was enabled and thought these functions might be a nice addition to the framework API:

- getEnabledModules([$pid])  
  Gets enabled modules on the system or for a project.
- isModuleEnabled($prefix [, $pid])  
  Returns true when the specified module is enabled on the system or the given project.
- isValidProjectId($pid [, $condition])  
  Check whether a project id is valid, performing additional checks with $condition (must exist or has a certain status).
- getProjectStatus([$pid])  
  Gets the status (DEV, PROD, AC, DONE) of the current (or given) project.


grezniczek:
You are right, of course, these should have been put into Framework.php. Done.

mmcev106:
Thanks much!

Issue #296, opened 2020-04-03
The addAutoNumberedRecord() method may not be safe for concurrent requests


mmcev106:
It attempts to be threadsafe by inserting then deleting possible duplicates, but that might just moves the "critical section" to the delete, making it possible to delete all newly inserted record ids.

123andy:
I was looking through the code around simultaneous survey entries in response to a post to the community site and felt as though the caching and lookup was not very efficient.  I think some form of db-lock is the only safe way to secure the next record id.  Either by locking a table for update or using the get_lock feature of mysql...

To add to this, there are many who want to 'customize' the syntax for addAutoNumberedRecord - such as "P001", "P002", ....   This would be an idea place for a hook.

mmcev106:
Are you envisioning something like a `redcap_get_next_auto_numbered_record_id()` hook?

mmcev106:
Perhaps for now we should simply remove our custom implementation and call save data with the undocumented auto ID flag set to true.

123andy:
Sorry - when I replied above - but I mean to say that the built-in redcap code being used for public survey record generation has a problem.  There have been multiple consortium reports of scenarios where two people posting simultaneously from a public survey end up getting the same ID where poster-2 overwrites the data of poster-1.  In one case, a pdf of data was sent to poster-1 that contained poster-2's data.

The code is complex due to the record caching as well.  It would be great if you could look at the source code and see if you and @taylorr4 could come up with a more atomic new record generator.  I think the only way to do it properly is a mutex in the database.  Perhaps for each project/arm there would be a 'last_auto_record_id'.  When someone calls getNextId - it would do a select FOR UPDATE to lock the row.  It would then use the current value to seed and begin checking if the next ID exists in the database.  If it doesn't, then it writes the new value to the table and returns.

If 10 people submit a public survey at the same time, they will have to wait in queue until the table becomes available - all of which would be managed by the innodb engine.

I would also love to see a hook point for this so we could easily add our own 'custom' record formatting.  People have complex requirements based on DAG, ARM, etc...

taylorr4:
[...]

PR #303, opened 2020-04-15
Improvement for addAutoNumberedRecord() for better accuracy

taylorr4:


classes/AbstractExternalModule.php
```diff
@@ -951,9 +951,11 @@ public function disableUserBasedSettingPermissions(){
...
 	public function addAutoNumberedRecord($pid = null){
 		$pid = $this->requireProjectId($pid);
-		$eventId = $this->getFirstEventId($pid);
-		$fieldName = \Records::getTablePK($pid);
-		$recordId = $this->getNextAutoNumberedRecordId($pid);
+		$Proj = new \Project($pid);
+		$eventId = $Proj->firstEventId;
+		$armId = $Proj->firstArmId;
+		$fieldName = $Proj->table_pk;
+		$recordId = \Records::addNewRecordToCache($pid, DataEntry::getAutoId($pid), $armId, $eventId);
```
mmcev106, line 958: With this change in place, I think perhaps the rest of the code in this method should be removed.  Do you concur?  If so, I don't mind being the one to actually remove it and test to make sure everything still works as expected.

You asked me to clean up the `addRecordToRecordListCache()` call below months ago, but I ran into some issues that I believe this commit might resolve.  Several things got added above that on my TODO list since then...  I apologize for not getting back to it sooner.

taylorr4: @mmcev106 no, the rest of the code in that method should stay. The addNewRecordToCache method merely sets a 1-hour hold on a potential record name. It does not actually create the record. (I know it's confusing because it has "cache" in the method name, but this was created long before the current "record list cache" in REDCap.) REDCap reserves the record name using that method, and then it creates the record (in various ways depending on the context). So you need to call that method and THEN create the record to make it a real record. So my code changes in this PR are fully correct and required to create a new auto-numbered record.

123andy: I just updated the post on #296 with a proposed solution that would work for most of my use cases...  I admit I don't fully grasp this change.  I agree that the naming is confusing.  I like my 'reserveNewRecordId' as a better name and I would have the 'addAutoNumberedRecord' do the loop and call it, similar to my two example code-bits...

mmcev106:
I'm going to go ahead and close this in favor of #306.  Feel free to re-open if anyone sees fit.

PR #306, opened 2020-04-15
Replaced the addAutoNumberedRecord() implementation

mmcev106:
This is mainly for @taylorr4 per our discussion on #296 , but any input is welcome.


mmcev106:
@taylorr4, I made a reminder to merge this to LTS at the same time it makes it into a standard REDCap release.

Issue #286, opened 2020-03-25
Appropriate for framework?  Utility class for working with repeating instances


123andy:
So, working with repeating instances in REDCap is challenge... They return fields that are part of the repeating form, you have to call a method on the new Project object to get it to initialize... Getting things like the 'next instance' is a bit of a challenge...

So, we started a helper class to work with repeat instances and I'm wondering, where, if anywhere, such a helper class belongs.

1) We can repeat it and namespace it to each EM that uses it.  but, this gets a bit repetitve.
2) We could try to make it part of Framework for EMs but that is maybe a bit overkill since it isn't itself a static method
3) We could invent a way to add classes to redcap, like a 'type' of external module... Not sure where that goes...

Just thought I'd bring it up for discussion...

grezniczek:
I have been asking myself the same question. Where best to put reusuable bits of code? Especially those that might benefit others in the community, too, so not everybody needs to reinvent the same wheels.

An idea was that such utility classes, functions, ... could be consolidated in the [External Module Templater](https://github.com/vanderbilt-redcap/external-module-templater) (EMT) GitHub repo (the deluxe-version would have integration with the GUI builder ... "Include XY library"). This would both, facilitate code-sharing (one doesn't necessarily have to use the EMT module) and prevent  version conflicts, as the utilities would always be scoped to the EM that uses them.

Alternatively, a separate (curated) GitHub repo could be established. Can we build a `redcap-compose`? Or just use composer and publish such classes with packagist. I've never done it, but it seems to be little effort? A naming convention could be established and a directory (just a README here on GitHub), that is linked to from the EM docs.

123andy:
Are you suggesting that when you make your EM template you could choose to include helper-library-x and it would pull that class into your template and namespace it?    So, your final EM would be a self-contained entity but you would get the benefit of having a library of helper classes? 

I like this idea!  Reminds me I've also make some modifications on the templater I need to push up...

grezniczek:
Yes, that would be the idea if the EMT is chosen as the host for stuff like this.

mmcev106:
This is something I've struggled with greatly... Even within my own team it's been very difficult to create a shared library that is more appealing that copy/pasting and reinventing the wheel.  I'm not entirely sure why...  The best success we've had so far is with the EM framework, but I still kind of feel like I'm on my own with it most of the time.  I also haven't had any incentive to make new module features available outside the context of a module since I always create new projects as modules rather than plugins...

[...]

PR #307, opened 2020-04-16
Get repeating events

grezniczek:
@mmcev106 
Here are some more contributions to the framework API:

- `getRepeatingFormsEvents([$pid])`  
   Gets repeating forms and events (including the forms on the event) for the current or specified project.
- `getEventForms([$event_id])`  
   Gets the forms in the current or specified event.

The benefit of `getRepeatingFormsEvents` over the existing `getRepeatingForms($event_id, $pid)` is that the former will not throw unexpectedly, while the latter will throw when called with $event_id = null and a $pid of a project with more than one event.


grezniczek:
Actually, before you consider these ... please give me some indication if you would. I also have another function I just finished in one of my EMs that might be useful: it gets a map of the complete data structure of a project, which is quite useful for understanding the data returned by e.g. REDCap::getData().

Before merging, I would move (most) of the logic to ExternalModules as static methods, so they can be used internally as well.

123andy:
@grezniczek  -  I would love to incorporate something like this into the framework as well but haven't had time.  I'd like to make one for repeating events as well.  

https://gist.github.com/123andy/e869115f69a6dc6a04651b40e0bc41c5

I'm not sure how that works with your proposed methods here.  But it should be easy to say:
->getLastInstance($form, $event)
->getNextInstanceID($form,$event)
->saveInstance($form_data, $instance_id, $event), etc...

grezniczek:
@123andy - you are way ahead of me (as usual ;)
I am still at a stage where I simply want to get the tools to get a better/unified representation of the data structure of a project (i.e. fields, forms, events, arms, what fields are where, and whether the forms/events they are on are repeating). Now, this information is spread around in differnent parts (a lot in $Proj, but nothing really regarding repeated stuff). My goal was to have this so it's easier to programmatically (based on some module configuration) pull out data from the right place. I find it incredibly tedious to work with the data structure returned by REDCap::getData :(

However, I haven't thought about (making) new framework API methods for retrieving data from or putting data back into REDCap. Yet. But it certainly would be nice to have tools that are more to the point than picking apart/assembling a complicated data structure for REDCap::getData and saveData.

I'll amend this PR with the method I made to get a complete representation of a project's data structure soon .. still need to document it better.

I also need/want something like this for EM config settings.

123andy:
I agree - the structure of the getData call for repeating forms and instruments is really hard to use.  We need some more friendly wrappers that can be exposed as plugin methods or EM methods.

grezniczek:
@mmcev106 , @123andy - 
I have moved the (static) implementations to ExternalModules and added another method:
- `getProjectDataStructure($pid = null)`

It gets a description of a project's data structure (see outline below). It would be very useful to build a helper class around this that provides the kind of methods suggested by @123andy.

Something like a `ProjectDataHelper` class that can be new'd up with a project id. It would use the project data structure internally as a basis for reasoning, and expose methods to retrieve or manipulate data. A separate `RecordDataHelper` class might be useful for working with data of a specific record.

```
[
  "forms" => [
     "form name" => [
         "name" => "form name",
         "repeating" => true|false,
         "repeating_event" => true|false,
         "arms" => [
             arm_id => [ 
                 "id" => arm_id 
             ], ...
         ],
         "events" => [
             event_id => [
                 "id" => event_id,
                 "name" => "event name",
                 "repeating" => true|false
             ], ...
         ],
         "fields" => [
             "field name", "field name", ...
         ]
     ], ...
  ],
  "events" => [
     event_id => [
         "id" => event_id,
         "name" => "event name",
         "repeating" => true|false,
         "arm" => arm_id,
         "forms" => [
             "form_name" => [
                 "name" => "form_name",
                 "repeating" => true|false
             ], ...
         ]
     ], ...
  ],
  "arms" => [
     arm_id => [
         "id" => arm_id
         "events" => [
             event_id => [
                 "id" => event_id,
                 "name" => "event name"
             ], ...
         ],
         "forms" => [
             "form name" => [
                 "name" => "form name"
             ], ...
         ]
     ], ...
  ],
  "fields" => [
     "field name" => [
         "name" => "field name",
         "form" => "form name",
         "repeating_form" => true|false,
         "repeating_event" => true|false,
         "events" => [
             event_id => [ 
                 (same as "events" => event_id -- see above)
             ], ...
         ],
         "metadata" => [
             (same as in $Proj)
         ]
     ], ...
  ]
] 
```

biggeeves:
These are great ideas.  I'd really like some place to grab the entire project's data structure.

I would be very careful about duplicating ways to get information in the data structure.  I wouldn't place the project structure within the field data.  Fields having the information for Form Name, repeating_form, events, etc is too much.

grezniczek:
@biggeeves - 
The idea is to make it simple to get to information easily no matter what your starting point is.
If you have a field name and want to know if its on a repeating event,  or want to get a list of events it's on, then you'll just love that the info is there and accessible with an easy to understand oneliner.

As this is a purely read-only structure I don't see any harm in duplicating information. What would be the drawback?

grezniczek:
But yeah .. the redundancies could be dropped as they would be unncessesary with the helper classes I mentioned. But we don't have these (yet).
They'd allow stuff like `getEventsByField($fieldname)`

123andy:
Here is a class I made earlier this week designed to get form data and handle both singleton and repeat instances.  The documentation needs a lot of work, but it is pretty simple.  Each form is just an array of data.

https://github.com/susom/redcap-select-repeat-instance/blob/master/FormHelper.php

Ideally it should work in repeating events as well, but still needs some testing...

```php
$f = new FormHelper($instrument, $event_id);

// Make sure the destination is a singleton event
if ($f->getEventType($destination_event_id) !== $f::TYPE_SINGLETON) {
    $this->emLog("$instrument is enabled as a non-repeating form in the destination event $destination_event_id");
    continue;
}

if ($f->getEventType($source_event_id) == $f::TYPE_SINGLETON) {
    // We are doing a normal copy of the data
    $this->emDebug("Doing a singleton save");
    $data = $f->getData($record, $source_event_id);
} else {
    $this->emDebug("Doing an instance save $repeat_instance data");
    $data = $f->getData($record, $source_event_id, $repeat_instance);
}

$this->emDebug("Save Data is ", $data);
$result = $f->saveData($record, $data, $destination_event_id);
```

biggeeves:
@grezniczek redundancies increase the effort needed maintain code.  From a database point of view, I wouldn't want a field to include the form name in REDCap.  This would be necessary if a single variable name could be used in two forms, but as of yet, REDCap doesn't allow that. The relationship is one form to many fields.  

It would be nice to get all of the field information in a single compact form and not have to make more calls.  I'm just not sure redundant is a good idea in a project structure.  A wrapper that delivered that data may be more maintainable down the road.

grezniczek:
`getRepeatingFormsEvents()` probably should be removed in favor of a "helper". 

How about consolidating the getters in this PR (and any new related ones, and maybe even a few of "old" ones) in the instance returned by `getProject()`, as you suggest, @mmcev106 .

This class instance returned by `getProject()` could use the data structure (lazy-generated only when a method on that instance is called that actually needs it) internally and offer wrappers for common use cases ((and also expose the raw structure on demand via `getDataStructure()`). Since the data structure is cached in the instance, any further calls to its methods would be cheap (i.e. no further DB access).

I wonder if `getEvent($eventId)` should be on there, too. As event ids are unique independent of project ids, this is not actually needed, but might be a good safeguard by throwing an exception when an eventid doesn't match up with the project.

grezniczek:
Please see my comment over here: https://github.com/vanderbilt/redcap-external-modules/issues/286#issuecomment-633305718

grezniczek:
Closing this in favor of #337

PR #309, opened 2020-04-17
Amendment to v5 version of getProjectSettings()

grezniczek:
This is an amendment to an earlier (merged) PR: #277 
If v5 modules would call `$this->getProjectSettings()` they would still get the old behavior.

This PR fixes the issue by checking the framework version and forwarding from the implementation in `AbstractExternalModule` to the one in `Framework`.

It was necessary to make the private $VERSION accessible through a new `getFrameworkVersion()` method.

I haven't checked, but similar issues may be present with other (breaking) changes. This should be watched out for with the new paradigm that framework methods can be called on "$this".
Or am I getting something wrong here?


classes/AbstractExternalModule.php
```diff
@@ -210,6 +210,10 @@ function getProjectSetting($key, $pid = null)
 	 */
 	function getProjectSettings($pid = null)
 	{
+		// Ensure that EMs using v5 get the v5 version.
+		if (isset($this->framework) && $this->framework->getFrameworkVersion() >= 5) {
+			return $this->framework->getProjectSettings($pid);
+		}
```
mmcev106, line 216: Hmm, this is a case I hadn't considered...  This is currently problematic for other functions like `getSubSettings()` as well.  I think we should try to prevent the need for this.  I think that best way to do that is to instantiate the framework class for v1 as well.  I can work on this, but I'd love your input.  I see two options:
- Go ahead and set `$module->framework` even for v1 modules.  Technically there's a risk of conflict here, but it may not be an issue for anyway.
- Move all `Framework` instances out of the module class to prevent possible conflicts.  We could store them in a static array indexed by prefix.  This would be pretty easy to do...

What are your thoughts?  Do you see any other ways to prevent the need for this duplicate framework version checking in each legacy method?

docs/framework/intro.md
```diff
@@ -57,6 +57,7 @@ getConfig() | ?.?.? | get the config for the current External Module; consists o
 getEnabledModules([$pid]) | ?.?.? | Returns an array with the modules enabled on the system or for the project with the given project id. The array is of the form "prefix" => "version".
 getEventId() | 9.7.6 | Returns the current event ID.  If an 'event_id' GET parameter is specified, it will be returned.  If not, and the project only has a single event, that event's ID will be returned.  If no 'event_id' GET parameter is specified and the project has multiple events, an exception will be thrown.
 getFieldLabel($fieldName) | ?.?.? | Returns the label for the specified field name.
+getFrameworkVersion() | ?.?.? | Returns the framework version number.
```
mmcev106, line 60: I have mixed feelings about exposing this method as a publicly supported method call.  Ideally the framework version handling would be internal, and module authors would never have a need to do any framework version checking themselves.  Leaving this in may encourage convoluted module design...  What do you think about removing it?

grezniczek: Agreed. I don't know why I documented it there. Pure automatism.
There is no need for it, as a module will always know its version ... after all, that's what they set in config.json, and thus is accessible with getConfig().

Removed with https://github.com/vanderbilt/redcap-external-modules/pull/309/commits/6db9459bf62ded8d99df099be4934f219888346d

classes/AbstractExternalModule.php
grezniczek: Do I understand the 2nd point correctly? Remove the "framework" part in `$this->framework`? That would break all v2+ modules that use $this->framework->xy(). Or is there a way to capture these calls? There seems to be no magic like `__call` covering this. 

But for v1 modules, a v1 framework instance could indeed be instantiated in a static array in ExternalModules with prefixes as key as you suggest. , and `__call` in `AbstractExternalModule` would need to check the framework version and use that "ouside" framework instance in case of v1.
This is quite straightforward, and it would avoid any risk. After all, "framework" is not that uncommon a term ...

grezniczek: Modifying __call will be the easy part. Moving stuff out to the v1 implementation seems more of a challenge ... all class variables will have to remain behind and replaced with $module->variable . Hmm, I am tired and may not be thinking straight, but it gets more complicated the longer I think about this ... v2+ calls that get sent back to AbstractExternalModule would again need to be forwarded to the v1 .... 

I wonder if the check as in this PR might be the easiest to apply to those (still few) cases were breaking changes between new frameworks and (existing and in number limited) v1 methods occur. And as they are breaking changes, they need special attention (documentation) anyway, so adding this guard is just part of this routine.

mmcev106: My 2nd point would definitely required a `__call` forward to wherever the new location is.  I think that's do-able though.  It'd be nice for be behavior to be identical for v1 as all other versions.  I'll try to throw at least an hour at this over the next few days and see how far I get.

grezniczek: A benefit of the version check added retroactively to v1 methods would be that this clearly states, in code (and not only docs) that a particular v1 implementation will not be used any more from the given version onward. And it avoids the bouncing through multiple `__call`s and confusion for devs debugging into framework code (I personally hate it).

grezniczek: > It'd be nice for be behavior to be identical for all versions (including v1). 

I am afraid v1 will always be special, because it's the framework without framework ;)

mmcev106: I think PR #319 might allow us to utilize the framework instance internally for v1 modules as well. What do you think?

grezniczek: Nice idea to flip them around. This certainly will harmonize things from the internal perspective of the framework.

I can't quite see (yet) how this would benefit the "user" (module authors) in regard to calling framework methods directly on the module instance. For those implemented directly on AbstractExternalModule and broken in a newer framework version will still need the explicit check shown above (#309). Am I missing something?

mmcev106: Since the general idea sounds good to you so far, I'll add another commit to prevent duplicate checking in each legacy method (though you might not like the `__call()` based implementation, heh).

mmcev106: PR #324 intends to replace this PR using an implementation that prevents the need for duplicate framework version checking, and makes the new `getSubSettings()` implementation the default in v5 (to prevent confusion about methods being forwarded from the module instance to the framework instance).  What do you think?

grezniczek: See my comment in #324. Looks fine otherwise.

mmcev106:
I'm going to go ahead and close this since I think #324 puts us in a better place for future changes.

PR #312, opened 2020-04-20
Expanded cron documentation

mmcev106:
I didn't think hard enough about a `cron_max_run_time` value and caused a module I was working on to crash.  After troubleshooting it, I figured I might as well update the docs on billable time as a reference for myself in the future.  Does this change make sense to y'all?


mmcev106:
After more thought, it's even more complex than this since one long running cron's run time get's rolled into the `cron_max_run_time` calculation for every other cron triggered in that same minute. 
 We just need to build better protection for long running crons into the framework...  I just created issue #313.  I'm not sure what else to do about it in the mean-time besides setting really high `cron_max_run_time` values.

mmcev106:
This is an improvement, despite it's imperfections.  Merging...

Issue #311, opened 2020-04-20
Saving system settings not working for modules enabled in projects


lsgs:
I've recently upgraded from 9.2.5 to 9.7.8 and I not get the exception below when saving system settings (e.g. to enable user activation permission) for any module that is enabled on one or more projects individually.

save-settings.php passes `null` as the project id to `ExternalModules::saveSettings()`, but in the way the query is built in `ExternalModules::getSettingsQuery()` it seems that this is unexpected - it looks like that method expects a value of `self::SYSTEM_SETTING_PROJECT_ID` rather than null, but that this is in the process of being deprecated.
```
An error occurred while saving settings:

<br />
<b>Fatal error</b>:  Uncaught Exception: More than one (32) External Module setting exists for prefix 'dag_switcher', project ID '', and key 'enabled'! This should never happen! in /site/redcap/redcap_v9.7.8/ExternalModules/classes/ExternalModules.php:2173
Stack trace:
#0 /site/redcap/redcap_v9.7.8/ExternalModules/classes/ExternalModules.php(1846): ExternalModules\ExternalModules::getSetting('dag_switcher', NULL, 'enabled')
#1 /site/redcap/redcap_v9.7.8/ExternalModules/classes/ExternalModules.php(301): ExternalModules\ExternalModules::setSetting('dag_switcher', NULL, 'enabled', false)
#2 /site/redcap/redcap_v9.7.8/ExternalModules/manager/ajax/save-settings.php(22): ExternalModules\ExternalModules::saveSettings('dag_switcher', NULL, Array)
#3 {main}
  thrown in <b>/site/redcap/redcap_v9.7.8/ExternalModules/classes/ExternalModules.php</b> on line <b>2173</b><br />
				<div class="red" style="margin:20px 0px;max-width:700px;">
					<b>REDCap crashed due to an unexpected PHP fatal error!</b><br><br>
					<b>Error message:</b> Uncaught Exception: More than one (32) External Module setting exists for prefix 'dag_switcher', project ID '', and key 'enabled'! This should never happen! in /site/redcap/redcap_v9.7.8/ExternalModules/classes/ExternalModules.php:2173
Stack trace:
#0 /site/redcap/redcap_v9.7.8/ExternalModules/classes/ExternalModules.php(1846): ExternalModules\ExternalModules::getSetting('dag_switcher', NULL, 'enabled')
[...]

PR #314, opened 2020-04-21
Fix save system settings

lsgs:
Handle empty/null pid parameter when saving module system settings. See [issue 311](https://github.com/vanderbilt/redcap-external-modules/issues/311).


mmcev106:
See #311 for more details.

grezniczek:
Well, @mmcev106 , the real problem is actually a bit further downstream at the conversion to an array. I had looked that up, but it was late and I didn't bother to make a note of it. Sorry.

This PR ensures only the normally expected behavior of php when getting an empty url parameter, namely that it results in an empty string (and not null) - see https://github.com/vanderbilt/redcap-external-modules/issues/311#issuecomment-617426915 and link therein.

An interesing (and yet unresolved) question is how in Luke's case the null got to happen in the first place, which seem to be rather special, else we would have seen this issue much earlier and frequently.

Issue #257, opened 2020-02-11
v9.6.5 EM not firing for some hooks if previous hooks resulted in an email going out


123andy:
We have a 'redcap_save_record' hook that stopped firing at 9.6.5.  We put in a bunch of debugs and got this far:

We have around 80 EMs on the server.
A particular project has 10 EMs enabled.
When we save a survey on this project, we are finding that a hook isn't working anymore.

1) When we get to the save_record hook, the External Module builds an array of all the EMs in `callHook` with `$versionsByPrefix`.  In our case we have 10 some EMs enabled.

2) It looks at each enabled EM for a save_record function and, in our 8th em finds one which has a saveData call in it.

3) When saveData is done, this triggers Alerts and Notifications going off, which triggers an outgoing email.

4) The outgoing email results in the hook 'redcap_email' being processed against ALL EMs on the server.   So, we were 8/10 through the save_record hook and now we start 1/80 on the redcap_email hook...

5) When the email hook is finished, it returns to complete the last two modules with 'save_record'.  However, a static variable inside the static ExternalModule class no longer has the right value for `self::$hookBeingExecuted` -- it is null.

So, it seems that if we keep a static class to run EMs, we need to retain the previous context before starting to process the nested email hook (or something like that..)

If we turn off alerts and notifications so that no emails are sent, then things work as expected and it processes all the hooks.  So, the injection of the redcap_email hook processing looses the information about the previous hook being run...

```
// HERE WE START PROCESSING THE REDCAP_SAVE_RECORD HOOK...
2020-02-11 14:54:59	865.9	15793	[survey respondent]	Hooks	39	call	[1/1]	STR	======= redcap_save_record =======
2020-02-11 14:54:59	866.6	15793	[survey respondent]	ExternalModules	2584	callHook	[1/1]	STR	save_record: 10
2020-02-11 14:54:59	866.7	15793	[survey respondent]	ExternalModules	2590	callHook	[1/1]	STR	 [1] checking stanford_webauth / v9.9.9 ...
2020-02-11 14:54:59	866.9	15793	[survey respondent]	ExternalModules	2590	callHook	[1/1]	STR	 [2] checking imagemap / v9.9.9 ...
2020-02-11 14:54:59	867.2	15793	[survey respondent]	ExternalModules	2590	callHook	[1/1]	STR	 [3] checking em_logger / v9.9.9 ...
[...]

PR #315, opened 2020-04-23
Added a call stack to better support nested hooks (like redcap_email)

mmcev106:
This aims to fix issue #257.


mmcev106:
If y'all could pay especially close attention to the `$delayed` call changes, I would very much appreciate it.  It was suspiciously already set up to support this well.....which makes me feel like I'm missing something.  I'd like to test that feature more thoroughly before merging this.  I know the `$delayedLastRun` flag will currently get mixed up between nested calls, but that gets into very uncommon territory.  I might be comfortable merging this as-is and getting the ball further down the field.

mmcev106:
I think I may have found an easy way to fixed the `$delayedLastRun` issue (see the latest commit).

mmcev106:
I am leaning toward closing this PR, and instead making a new one that uses @123andy's [original fix](https://github.com/vanderbilt/redcap-external-modules/issues/257#issuecomment-584917587) plus commit 7fbe33e.  The call stuck just isn't benefiting us at all (yet), and it's a risky change...

mmcev106:
Closing per https://github.com/vanderbilt/redcap-external-modules/issues/257#issuecomment-622562848

PR #316, opened 2020-04-24
Populated dropdown settings only if getting the config for the settings pages

mmcev106:
Commit 86358f5 inadvertently caused dropdown options for generated settings fields (project, form, field, user, etc.) to get added to the config in contexts other than displaying the settings dialogs.  This was already happening in contexts where `$_GET['pid']` was set, but that commit expanded it to all cases and caused the Email Alerts modules to crash a few times (oddly, only on survey queue pages).  This PR moves that action to what I think is a more appropriate location.  We call `getConfig()`  all over the place, but I think the new location is the only place where we really need to auto-populate those dropdown options.


mmcev106:
I'm going to go ahead and merge this so it can be tested on redcaptest.  @kcmcg, I'd still like to wait on your review before merging this to production.

kcmcg:
This looks good. It's always possible to manually pull the field settings should it be needed.

PR #320, opened 2020-04-29
Optimize config translation

grezniczek:
This PR brings some optimizations to the module translation feature.
- Greatly reduce the number of database requests by loading module language settings only once per context (system or project); query results are cached.
- Optimize translation of config.json. This will now happen only once, and not repeatedly with every getConfig() call. Translations are cached. 

I don't belive there to be any risk in caching either of the two, as config.json will likely not change in the middle of a request (nor should this ever be a use case).


mmcev106:
Merged by proxy of #321

PR #321, opened 2020-04-29
Optimize config translation

mmcev106:
This is an extension/remix of [this PR](https://github.com/vanderbilt/redcap-external-modules/pull/320).  The concepts are the same, but I made a few changes in the interest of simplication/maintainability.  What do you think @grezniczek?


classes/ExternalModules.php
```diff
@@ -3424,7 +3443,8 @@ static function getConfig($prefix, $version = null, $pid = null, $translate = tr
 			$version = self::getEnabledVersion($prefix);
 		}
 
-		$config = @self::$configs[$prefix][$version];
+		// Is the desired configuration in the cache?
+		$config = @self::$configs[$prefix][$version][$translate];
```
grezniczek, line 3447: Maybe use a getter function here, matching the setter (`self::cacheConfig()`)?
`$config = self::cachedConfig($prefix, $version, $translate)`

grezniczek:
Sure, looks good.
Only thing maybe worth changing .. see comment in code.

grezniczek:
I see. Ok. Never mind the getter then. 

This could be further optimized. When a translated version is requested, the non-translated should always be added to the cache, as it's "free".
If a translated version is requested and the non-translated is already in the cache, then use this instead of going to the disk.

Want me to make a PR for this?

mmcev106:
I actually intentionally cached them separately (see [this comment](https://github.com/vanderbilt/redcap-external-modules/pull/321#issuecomment-621892169)). I was thinking it was worth avoiding the unnecessary queries and other actions on every request.  I know `fillModuleLanguageSettingCache()` is only called once per request now, but there are many cases where it doesn't need to get called at all (cases where no module hooks end up firing).

mmcev106:
I still think we likely should change `$translate` to default to `false` for this reason and change the few cases where we actually need it to pass `true`.  Once step at a time though...

grezniczek:
Yes. 
I've just looked through the ~31 places getConfig is called. I think a change of the default value for translate is warranted after all. In the (probably) rare circumstance that a call with $translate = true happens first, the non-translated version should be cached in any case. 
For requests for translated configs, a cached non-translated should be used for translation, if it exists.

As is now, duplicate trips to disk/db could happen.

Oh - when I started typing this (was interrupted) your https://github.com/vanderbilt/redcap-external-modules/pull/321#issuecomment-621951201 hadn't shown yet.
So we agree. Want me make a PR?

mmcev106:
If you're able, a PR would be very much appreciated.  If you have any other thoughts after looking into it more, feel free to share them.

PR #324, opened 2020-05-05
V5 simplication

mmcev106:
This is primarily for @grezniczek and is intended to replace #309 and prevent the need for duplicate framework version checking.


grezniczek:
@mmcev106 , what are the reasons for having a public `getSubSettings_v1()`? 
The name change would require module authors to explicitly use this (i.e. change their code). Instead, if they want their module to be v5 or newer, they should rather adapt to the new return format, which, if I am not mistaken, on the first level is equivalent anyways. 
Thus, this v1 method could be private and eliminated from the docs (stating the change with v5 is enough).

mmcev106:
I just added another commit that was necessary to make all cases work, but also simplifies some things (and makes it easier to stay simple going forward).

There may be modules (even here at Vandy) that might need to update to v5 but do not have the budget to figure out how to automatically refactor/resave complex nested settings that use the old sub-settings implementation.  We already have the `getSubSettings_v1` method for backward compatibility, so it seemed a little cruel not to allow access to it for projects that might be in a hard spot budget-wise.

The main goal was to make the new implementation the default, regardless of how it's called.

grezniczek:
Fair enough.

I've just been thinking ... why not make this a general thing?
If a method is called by its base name, e.g. `getSubSettings()`, the most recent implementation (according to the version requested by a module in config.json) is used. For every method, a version suffix is supported (such as in `getSubSettings_v1()` or `getSafePath_v4()`) that will call the appropriate version (i.e. the one that would have applied when calling the base name and config.json specifying this version). Implementation should be easy ... simply replace VERSION with that extracted from the suffix in the decision making.
If I am not mistaken, this approach would even make having different framework version instances obsolete. All that needs to be maintained is a mapper for basename, version, and actual implementation (entries only for when breaking changes occur, the correct version to call is the one specified if it exists or the next lower). Version is just a number, that can change on the fly on a call by call basis.

Again ... just an idea. I may have missed some crucial issues. But I think this could work without much effort.

mmcev106:
Those are interesting ideas.  I may need to know more implementation specifics to have a strong opinion.

mmcev106:
I went ahead and merged this to testing so it could be running on our test server.  Please feel free to continue any discussion here if anyone has any thoughts/concerns.

grezniczek:
> Those are interesting ideas. I may need to know more implementation specifics to have a strong opinion.

I'll see if I can come up with some demo code. It may take a while ...

PR #331, opened 2020-05-28
Project sub setting fix

mmcev106:
@tmwil, I believe this solves the Cross Project Piping issue we were talking about during code review for all cases.  Can you test this and verify that everything works as expected?  Once you approve, this should be safe to merge & deploy to production.


mmcev106:
@tmwil, I just added a commit.  You likely should re-review & re-test to make sure this changes still works as expected for you.  It has already been merged & deployed to testing if you want to test there.

mmcev106:
Thanks @tmwil.  Merged!

PR #336, opened 2020-06-18
HTML in error messages was escaped

grezniczek:
Found a location where translations should be passed through `tt_raw()` to preserve HTML tags.
I changed all occurrences here, as these strings are all directly output, and future changes to the text would not cause this issue again. Currently, only the last error (exception capture) has HTML tags in it.


mmcev106:
@grezniczek, would you mind going into a little more detail as to why this change is proposed?  Everytime we use `tt_raw()`, we introduce a javascript injection vulnerability.  If possible, I think we should consider alternatives.  Also, since `em_manage_90` doesn't have any parameters, `tt_raw()` wouldn't make a difference would it?

Thanks

grezniczek:
Sure, @mmcev106. 
Some exception messages contain HTML, e.g. ExternalModules.php, line 3019.  This is one that could end up being shown in line 60 of this PR's file.

```php
if (!empty($Exceptions)) {
	//= COMPATIBILITY ERROR: This version of the module '{0}' is not 
	// compatible with your current version of PHP and/or REDCap, so cannot be
	// installed on your REDCap server at this time. Details:
	// Remove any potential HTML tags from name for use in error messages.
	throw new Exception(self::tt("em_errors_43", strip_tags($config['name'])) . 
	"<ul><li>" . implode("</li><li>", $Exceptions) . "</li></ul>");
}
```

I saw escaped tags in an error I got - this was the "trigger" for this PR.

Then the thought process was that there is no harm in changing the others, as 
- the source (English.ini) must be considered safe. Whoever could introduce malicious code here could do it in REDCap's strings as well, and for the most part, there is no filtering whatsover.
- the others pass $module_name, and there is no rule that would prevent a module author from using HTML in a module's name. The question here would be: Should this be officially disallowed? Currently, this works and is displayed fine on the module management page, but not in these error messages. And if a module author wanted to be malicious ... well, there are better (or rather much worse) ways.

An alternative would be to strip tags from `$module_name`, then `tt()` would be fine for those cases. 

**And now that I say this, I realize, that this is what actually already happens** ... :[  I guess someone (probably me?) already thought of this (and then forgot; and did not check).

So, I reverted back to `tt()`. https://github.com/vanderbilt/redcap-external-modules/pull/336/commits/a990d8094cf5aeb054a6862b0c72b7981cefb4c3

But the error message remains a valid case for `tt_raw()`.

mmcev106:
What is an example of a case where `$exception->getMessage()` should not be html escaped?  One of the cases we specifically considered when we decided to escape tt params was exception messages, since they sometimes contain POSTed input or values from the database (both of which can be injected with malicious intent).

grezniczek:
Well, the one in my post. The code block is copy/paste from ExternalModules.php. It specifically adds HTML to the output. 

Maybe `Exception` should not be used to "transport" error messages intended for display to the user?

mmcev106:
It's definitely a sensitive area...  Perhaps we the tt functionality should generally be limited to only language (not structure like HTML).  We can always echo HTML after any tt strings, with appropriate html encoding on the content of any desired HTML tags.

grezniczek:
I don't think this is feasible. 

Just look at English.ini (take your pick: REDCap's in LanguageUpdater, or the Framework's in classes). Lot's of HTML tags (mostly **emphasis**, bold and italics, paragraphs, and line breaks.

Emphasis is something that **must** be "translatable" (i.e. contained in the language strings), as this **cannot** be done as an afterthought. 

I stand by my argument made above: Language files (or config.json for that matter) **must** be considered a safe source. Else one would have to give up _any_ (flexibility in) formatting whatsoever. That would result in very bad UX.

mmcev106:
That makes sense.  I think things only become problematic when we introduce dynamically generated content (like error messages).

grezniczek:
The problem is: How to discriminate? That is: between safe and potentially unsafe when exceptions are used to transport display text.

Anyway, in the one example given above, the output will always be safe (as in: all that is contained in it comes from defined sources), but there are a **lot** of potential exceptions that could end being displayed here. However, as far as I could see in a rather superficial check, the only potential dangerous source that might come into play is config.json. 

And then again ... why (should the person with malicious intent) bother with config.json, when there are PHP files to mess with?

mmcev106:
One of the main cases that lead to escaping language params & errors was due to javascript being injected into an error message via a module prefix parameter.  I think we can and should think harder in order to appropriately discriminate to achieve all desired goals while also avoiding vulnerabilities.  In the `em_errors_43` example, it might be as simple as removing the language parameter and manually appending adequately escaped HTML as follows:
```
$errorDetails = '<ul>';

foreach($Exceptions as $exception){
    $errorDetails .= '<li>' . htmlspecialchars($exception) . '</li>';
}

$errorDetails .= '</ul>';

throw new Exception(self::tt("em_errors_43")) . $errorDetails);
```

grezniczek:
Ok, for the one case here, not using interpolation will work fine. See https://github.com/vanderbilt/redcap-external-modules/pull/336/commits/a3bf60e1ed90c25cdb19e5e5c8b44a2154a69b81. This eliminates `tt_raw()` in favor of dot concatenation. I put an extra space in there for good measure, as the one after the colon in `English.ini` might be lost (in translation) eventually.

I am not sure I understand your example in this context, as here, nothing comes from a potentially unsafe source (all exculsively from `English.ini` and parameters are passed through `tt()`, i.e. they are escaped). But I agree that `htmlspecialchars` should be used to sanitize anything from a potentially unsafe source before being output.

I guess with the last commit, the problem this PR addressed (output of escaped HTML where it should not) is solved.

I am wondering though, how could JS be injected via prefix? I am not 100% sure, but it would appear that prefix can only enter the framework from 2 sources ... from a directory name, and form a download from the EM Repo. These two "entry points" could be secured by allowing only `[a-z]+[a-z0-9_]` (and the version bit for parsing directory names). Of these two, however, only the download is to be considered as potentially unsafe (as with file level access on the server, well, nothing much matters anymore anyways).

mmcev106:
I'm sorry, I merged this prematurely.  It has been reverted via #343.  The exception message is still not HTML escaped.  We need to escape all exception messages because they could now or in the future be modified to contain input that could be injected by someone with malicious intent.

I don't know if there is a current vulnerability in this case, but many of the vulnerabilities our security scanners have found are NOT obvious when reviewing the code, and were prevented by simply enforcing better standards like escaping.  I don’t think we can justify NOT following best security practices.

grezniczek:
I get that. In this case, however, exceptions **MUST** not be used to transport formatted message, throughout the framework.

This means ExternalModules.php, line 3019 (see above) needs to be fixed to NOT contain any HTML.

grezniczek:
Or if there are occasions where aggregate exceptions are useful, a derived `FrameworkException` class should be introduced that has a "Messages" property (`string[]`) that then can be output nicely formatted as a list (at the point where the output actually happens).

mmcev106:
What happens that is undesirable if `$return_data['error_message']` is HTML escaped?

grezniczek:
See the code snippet from ExternalModules.php in https://github.com/vanderbilt/redcap-external-modules/pull/336#issuecomment-653628812

You get the escaped HTML output on the screen, and not the bullet list.

mmcev106:
Does this code snippet solve that?  If not why?  
https://github.com/vanderbilt/redcap-external-modules/pull/336#issuecomment-653661027

Perhaps speaking via code examples might more easily bridge our gap and identify other solutions.

grezniczek:
No, this will not solve the problem, as HTML is part of the exception message that is then, in `manager/ajax/enable-module.php`, again passed through `tt()`, which escapes the HTML. 

If `$exception->getMessage()` is to be always escaped, then **no HTML must ever** be (purposefully or otherwise) be introduced into exception messages (this is totally irrespective of `tt()`).

The solution would be to not use any HTML in the message of the exception that is thrown in e.g. `isCompatibleWithREDCapPHP()` or find another way to pass aggregate information down exceptions (to be formatted at the output only).

Best to see the consequences of the current situation by trying to enable any module whose version (REDCap or PHP) requirements are not fulfilled.

mmcev106:
Based on past security scanner results, I think avoiding HTML content in exception messages might be a worthwhile trade off for the safety added by escaping exception messages.  I can see a few ways of separating the HTML generation from exception message content in this case (throwing individual exceptions instead of an array, subclassing exception to store an an array of exceptions, etc.).  What do you see as our next best alternative?

grezniczek:
I think the best (read: easiest, least work required) approach might be to throw "atomic" exceptions (one error at a time), at the cost of, in cases like the one that triggered this, it potentially taking several attempts to get the input fixed.
Or here: To simply forego the pretty list and just concatenate the messages.

Subclassing exception might me most elegant, but represents the most amount of work (and would probably still be very limited).

For now, I think #344 will do. 

Good news: I did a quick check (search and quick look at surrounding code) - this seems to have been the only place where HTML was used (at least in ExternalModules.php and AbstractExternalModule.php).

mmcev106:
I just merged #344.  I think that's a great compromise.  We can always come up with a fancier solution if we encounter similar needs in more places.  Thanks very much for baring with me on this.  I think it's important.

PR #337, opened 2020-06-22
Additions to the framework's Project and Record classes

grezniczek:
@mmcev106 @123andy 

We have previously talked about the need for some wrapper/utility classes that facilitate handling of REDCap projects and records (and their data).

This PR consolidates some "byproducts" of working on some modules that needed to manipulate records and their data, specifically repeat instances, in ways that are not supported by any existing APIs (be it calls to `/api/`, `REDCap` static functions, or even using calls into REDCap core code. I found that I had to write a lot of helper methods, and I tried to package them in a hopefully sensible and reusable way. 

**This PR is still a work in progress**. It should, for now, primarily show what I have come up with so far and serve as a basis for further discussion, and to consolidate more code some have probably already written.

The _Project_ class is a collection of methods that return information about various aspects of a project, such as it's features, the data structure, etc. Something that definitely could/should be integrated here is Andy's ActionTagHelper class.

This class is a beefed up version of the original framework class of this name that is returned by the `getProject()` method.

The _Record_ class is supposed to represent a single record in a project. It is instantiated via a _Project_ instance's `getRecord($record_id)` method and contains a set of tools to get info about and/or manipulate this record (such as lock state).

What do you think of this? Is this something that could be considered for a merge? 
I tried to name the methods in a consistent manner and follow common pattern. Does it make sense? What needs to be changed?
Anything that should **not** be in here?
As mentioned above - this is far from complete. I guess the list of what could be added is looooong.

As already discussed with Mark, care should be taken to minimize duplication of REDCap core code, and if core functionality needs to be deconstructed and reconstructed (the form deletion functionality is probably a very good example, as this is quite entangled with UI code in REDCap), then maybe such reconstructed code should get back into REDCap core (and used there to dis-entangle presentation and business logic). I guess, however, this won't be easy ...

I am not sure how to best proceed from here. Maybe this could be merged (soon),  so it's easier to work on this collaboratively? Documentation (i.e. "going public") could be held off for a while to let this mature.


grezniczek:
See also discussion in #307.

classes/framework/Project.php
```diff
@@ -1,30 +1,1270 @@
...
+        // Get REDCap's Project instance.
+        if (isset($GLOBALS["Proj"]) && $GLOBALS["Proj"]->project_id === $this->project_id) {
+            $this->proj = $GLOBALS["Proj"];
+        }
+        else {
+            $this->proj = new REDCap_Project($project_id);
+        }
+        $this->proj->getUniqueEventNames();
+        // Set permissions based on the current user. Let the module override as necessary.
+        $this->grantUserPermissions(USERID);
```
mmcev106, line 56: I'm hesitant to merge this new constructor code since any bug in it (as currently written) could prevent the constructor from succeeding and affect existing code.  I think it might be appropriate to make all this initialization lazy via getters (perhaps something like `getREDCapProject()` and `getPermissions()`, rather than referencing those properties directly).

grezniczek: I have now removed that logic from the constructor and replaced `$this->proj` with a (private) getter.
I am not initializing permissions at all, so they are all "off" by default and will have to be set explicitly by the module author. This is probably cleaner (at least more explicit) than to assume that the user rights of the current user should apply. This fact should be stated in the docs (eventually).

grezniczek:
@mmcev106, I had considered this, but then chose to construct the SQL as done throughout REDCap Core. There were two reasons for that:
1. All SQL statements are logged using `Logging::logEvent` - unfortunately, the parameterized query syntax does not give you the actually executed SQL statements. I think that logging is essential when modifying data stored in REDCap projects.  There would be the possibility to create SQL statements for logging separately, but then these have to be kept in sync (extra effort/burden/risk) and for any number of reasons (bugs, etc.) might not match what was actually executed.
1. Building them "classic" style is much more readable in many cases.

Where no logging is involved, I did use the parameterized queries (but these are all cases of SELECT queries).

I did look into getting the SQL that is actually executed from parameterized queries, but it appears there just is no way to get that.

Furthermore, the class internals are laid out and implemented in a way that will guarantee that all parameters are of a safe type: mostly they are integers such as instance number, event/arm ids (ensured by the parameter checks of each function) and record id (ensured by $db_record, escaped by `db_escape`), sometimes field names (which are escaped using `db_escape `or `prep_implode`).

However, logging was the main concern. Do you see a good solution for logging, other than to create Logging-SQL in parallel to the parameterized queries? Or is this just a prize/risk (more effort / risk of not being in sync) that **has** to be paid?

mmcev106:
As I think about this more I wonder if we should just avoid copy/pasting blocks of code from REDCap core going forward...  This has caused support issues in the past when the implementation in REDCap core changes.  How many of the logging queries would be non-issues if we extracted the required code to a method in REDCap core, then called that method from the framework instead of duplicating the code (including logging)?  I realize this is a tall order to do all at once, but we could do it granularly over time.

grezniczek:
Oh, believe me, nothing here is copy/pasted! If it were that easy! 
This was a huge effort of deconstructing complex code where UI was entangled with logic, reconstructing it as usable bits of functionality!!

;)

But I agree, and we had discussed this in our call a while back. Most of the code in the Project and Record classes should actually be in REDCap Core and only used by the Framework (as a publicly documented wrapper around higher order functional concepts such as getting info about a project, or modifying data, or deleting/creating an instance of an instrument in redcap_data).

mmcev106:
I am concerned that your effort toward disentangling all that functionality will go to waste if it doesn't make it back into REDCap core.  In the past, we've had to ditch framework implementations due to problems caused by drift...  Should this PR really be a REDCap core PR, with a thin wrapper for the module framework?

Perhaps we should consider a specific case.  Line 517 is in `deleteForm()`, which has a lot going on that looks like it's an high risk of drift from REDCap core.  What would it take to move that into REDCap core?  How much would be covered by moving the content of the `submit-btn-deleteform` `elseif` block in `DataEntry/index.php` into a shared method?

grezniczek:
Yes, of course this would be the best solution.

In this particular example, this would be rather easy. Lines 227-318 of `DataEntry/index.php` could more or less be extracted to a method as is (only a few parameters from `$_GET` need to be supplied as arguments, alongside `$Proj`). Of course, this would mean to use the extracted method in `DataEntry/index.php` as well. 

I think I remember it to be a bit more complicated in other cases, which would mean more extensive refactoring of REDCap code after extraction of functional blocks.

Would @taylorr4 be ok with such an approach? There is not zero risk for introducing (subtle) bugs, perhaps by _lack of side effects_ that are not obvious.

Where would such methods live? Probably in `Classes/Records.php`, or `Classes/Project.php`, as appropriate?

taylorr4:
@grezniczek yes, i'd be open to that being refactored and moving it out of DataEntry/index.php. Probably the best place for it would be Classes/Records.php.

mmcev106:
Thank you @taylorr4.  I could help with unit testing to improve confidence in changes, and reviews prior to things going to @taylorr4 if that's helpful.

taylorr4:
@mmcev106 sounds good to me.

grezniczek:
Great. I'll have a look a this and probably can make first PRs for the easy ones soon. This will also eliminate probably most if not all of the db_query calls the test runner complained about as listed in https://github.com/vanderbilt/redcap-external-modules/pull/337#pullrequestreview-446549847.

PR #340, opened 2020-07-03
Prevented updates for displaying for deleted modules

### LINKED ISSUE #338: external_modules_updates_available retains updates for deleted modules
REDCap 9.5.32, PHP 7.4.6

Through strange circumstances [1] we ended up with six external modules installed on our production server that we did not want. REDCap found that two of these had updates available. I deleted all six external modules, but REDCap is still prompting us to apply these two updates for external modules that no longer exist on the system. 

Value of external_modules_updates_available in redcap_config: {"729":{"name":"redcap_entity","version":"2.3.5","title":"REDCap Entity"},"744":{"name":"shazam","version":"1.3.8","title":"Shazam"}}

[1] We use separate web and database servers. In testing a planned migration from an old web server to a new web server, the new web server was connected to one of our staging databases. That staging database had the six external modules installed. REDCap apparently saw database entries for those external modules but no matching directories on the filesystem, so retrieved the six external modules and copied them to the file system. Later, after the migration of the web server, the new web server was connected to our production database. REDCap apparently saw the six external modules in the filesystem and added entries for the external modules into the database. REDCap then found that two of the external modules had updates available, and so added those updates to the external_modules_updates_available configuration.

mmcev106:
In theory this is a simple change, but I figured the areas touched were sensitive enough that it warranted review.  This aims to fix #338.


mmcev106:
I'll take two out or four :)

PR #345, opened 2020-07-07
Add support for nested hook calls

mmcev106:
Since the `redcap_email` hook was added, it is possible for hooks to be called from within other hooks. This currently overwrites some static variables and skips hook calls in some edge cases (see #257).  If correctly implemented, this PR should fix that problem while preserving all other existing behavior.  I've encapsulated all hook state into a `HookRunner` class that is instantiated per hook call, and reset to the previous instance to restore previous state on nested calls.  I did not move `$activeModulePrefix` to `HookRunner` because it is also used outside the context of hooks, but I did implement similar previous prefix restoration to (hopefully) support nested hooks in all cases.

This should also lay the groundwork for safely calling the save hook from within other save hooks.  I think we've experimented with this in a few places, but not yet officially supported it.

It may be easiest to review these changes on a commit by commit basis.  What are your thoughts?


mmcev106:
Since the hook state is saved and restored prior to each new hook call, the current implementation is not limited to two levels, and should currently support much deeper nesting (until we run out of memory I suppose).

mmcev106:
@123andy & @grezniczek, I believe this is ready to merged, but I would love your stamp of approval if possible since both of you weighed in heavily on the discussion on #257 that lead to this change.  Are either of you able to review this per chance?

classes/AbstractExternalModule.php
```diff
@@ -597,7 +597,7 @@ private function detectInstanceId($instanceId=null)
 	#		return;       // the module will be restarted from the beginning
 	#	}
 	public function delayModuleExecution() {
-		return ExternalModules::delayModuleExecution();
+		return ExternalModules::delayModuleExecution($this->PREFIX, $this->VERSION);
```
123andy, line 600: I've not considered delay module execution and nested hooks...  Do delays get 'queued' to the current recursion level or are delays global?  I'm not sure I can wrap my mind around this one after 5pm...

mmcev106: Delayed module info is stored at the current recursion level (in each instance of the HookRunner class).  I'm glad you asked since those var definitions were incorrect.  I just fixed them via [24b95b](https://github.com/vanderbilt/redcap-external-modules/pull/345/commits/24b95b749f119892fbd62fcd76fc2ab1574cc647).

PR #349, opened 2020-07-20
New getProjectTitle Function

knil-maloon:


classes/framework/Project.php
```diff
@@ -3,28 +3,42 @@
...
-		$results = $this->framework->query("
+        $results = $this->framework->query("
+                    select app_title 
+                    from redcap_projects 
+                    where project_id = ? 
+                    limit 1",[$this->project_id]);
+
+        while ($row = $results->fetch_assoc()) {
+            $this->project_title = $row['app_title'];
+        }
```
mmcev106, line 18: I don't think we want to perform any method specific actions in constructors.  The constructor may get called many times in a loop without the related method even being used.  I think we should use the lazy initialization programming pattern instead.  Also, this does not reuse the `redcap_projects` query that currently exists for project statuses like we discussed via slack.  We need to get into the habit of taking care of technical debt as we go for the module framework to be sustainable.  Something like to following might be the bare minimum while not introducing any more technical debt:

```
class ExternalModules{
    function getProjectDetails(){
        $results = $this->framework->query("
            select * 
            from redcap_projects 
            where project_id = ? 
        ", $this->project_id);

        return $results->fetch_assoc();
    }

    function getProjectStatus(){
        return self::getProjectDetails()['status'];
    }
}

class Project{
    private function getProjectDetails(){
        if(!isset($this->project_details)){
            $this->project_details = ExternalModules::getProjectDetails();
        }

      return $this->project_details;
    }

    function getTitle(){
        return $this->project_details['app_title'];
    }
}
```

What are your thoughts?

mmcev106: If you feel like also adding a `getStatus()` method and removing the docs for the old `getProjectStatus()` method, that would be easy and would improve consistency.  I won't ask you to do that unless you want to though.  We'd likely want to leave the old method in place for now, and maybe adding a comment in the code saying it used to be documented and could affect existing modules if removed.

knil-maloon: Oh Sorry! I completely forgot about the getStatus thing!
Also wouldn't it make more sense to add the pid and also add the status options as the old one?

f```
unction getProjectDetails($pid){
	    if(empty($pid)){
	        $pid = $this->project_id;
        }

        $results = $this->framework->query("
            select * 
            from redcap_projects 
            where project_id = ? 
        ", $pid);

        return $results->fetch_assoc();
    }

    /**
     * Gets the status of a project.
     *
     * Status can be one of the following:
     * - DEV: Development mode
     * - PROD: Production mode
     * - AC: Analysis/Cleanup mode
     * - DONE: Completed
     *
     * @param int|string $pid The project id.
     * @return null|string The status of the project. If the project does not exist, NULL is returned.
     * @throws InvalidArgumentException
     */
    function getProjectStatus(){
        $status = null;
        switch (self::getProjectDetails()['status']) {
            case 0: $status = "DEV"; break;
            case 1: $status = "PROD"; break;
            case 2: $status = empty($row["completed_time"]) ? "AC" : "DONE"; break;
        }
        return $status;
    }
```

Comment (knil-maloon) [reply classes/framework/Project.php:18]: Don't mind the $row['completed_time'] that would be changed with the new logic

Comment (mmcev106) [reply classes/framework/Project.php:18]: The `getProject()` method already supports a `$pid` param, so that would automatically apply to any method in the `Project` object.  Man, that status method is messy now that I look at it...  That raises a lot of questions...  Maybe we should only worry about reusing the query for now, and not worry about adding a `getStatus()` method just yet.

Comment (mmcev106) [reply classes/framework/Project.php:18]: ...and I just wrapped my head around what you suggested with the `$pid` param for `getProjectDetails()`.  I think that makes a lot of sense.

Comment (knil-maloon) [reply classes/framework/Project.php:18]: Do you want me to keep the old status then? I think the new change wouldn't affect how it works right now.

Comment (mmcev106) [reply classes/framework/Project.php:18]: The only status related change I think we need is a way to reuse the `redcap_projects` table query.

Comment (knil-maloon) [reply classes/framework/Project.php]: I added the new code. Let me know if you have any concerns

**Comment by grezniczek** [comment]
You all might consider #337. I haven't gotten around to rework the queries yet, or rather get get some of the functionality back into REDCap core and only provide a wrapper here, but it would seem that the framework's Project object, to provide the desired functionality, should really wrap a full REDCap Core Project object. This would also eliminate these SELECT queries.

#337 provides a `getStatus()`, but as of now only would return the numerical state. I guess it should provide a string (PROD, DEV, ...)

File: classes/framework/Project.php [hunk 2]
```diff
@@ -3,28 +3,41 @@
...
 class Project
 {
-	function __construct($framework, $project_id){
-		$this->framework = $framework;
-		$this->project_id = $framework->requireInteger($project_id);
-	}
+    function __construct($framework, $project_id){
+        $this->framework = $framework;
+        $this->project_id = $framework->requireInteger($project_id);
+        $this->project_details = $framework->getProjectDetails($project_id);
```
mmcev106, line 9: https://github.com/vanderbilt/redcap-external-modules/pull/349#discussion_r457579169

mmcev106:
@grezniczek, thanks for chipping in.  We did consider #337 and were a little overwhelmed.  I think we have to run with a simpler solution for now so that @knil-maloon can get the functionality she needs without a large time investment.

grezniczek:
Sure. Still, maybe consider a `new \Project($pid)` over running separate SQL. 
As I've learned recently, `new Project()` is rather cheap, as in most cases this will be the project that is already instantiated by REDCap, and thus only a copy of a cached version has to be made (and not a trip to the DB).

mmcev106:
Thanks @grezniczek, that does sound like a good approach.  @knil-maloon, it's up to you whether you want to get into that for this PR or not.  I'm planning on writing a unit test for `getProjectStatus()` before merging this.  As long as that passes, and no technical debt is added, I'm happy.

classes/framework/Project.php
knil-maloon: I made a commit using \Project instead. Let me know if it's not what you had on mind

classes/ExternalModules.php
```diff
@@ -5997,36 +5997,40 @@ public static function isValidProjectId($pid, $condition = false) {
...
+     * - PROD: Production mode
+     * - AC: Analysis/Cleanup mode
+     * - DONE: Completed
+     *
+     * @param int|string $pid The project id.
+     * @return null|string The status of the project. If the project does not exist, NULL is returned.
+     * @throws InvalidArgumentException
+     */
+
+    function getProjectStatus($pid){
```
mmcev106, line 6022: I think changing this method from static to non-static will break a few calls to it...

classes/ExternalModules.php
```diff
@@ -5997,36 +5997,40 @@ public static function isValidProjectId($pid, $condition = false) {
...
+        if (!self::isValidProjectId($pid)) {
+            throw new InvalidArgumentException(self::tt("em_errors_131")); //= Invalid value for project id!
+        }
+        $status = null;
+        switch (self::getProjectDetails($pid)->project['status']) {
```
mmcev106, line 6027: I believe this will now throw an exception on a missing project where it previously returned null.  Is this a save change for all calls to `getProjectStatus()`?

classes/ExternalModules.php
```diff
@@ -5997,36 +5997,40 @@ public static function isValidProjectId($pid, $condition = false) {
...
+            case 0: $status = "DEV"; break;
+            case 1: $status = "PROD"; break;
+            case 2: $status = empty(self::getProjectDetails($pid)->project["completed_time"]) ? "AC" : "DONE"; break;
```
mmcev106, line 6030: I don't think we should duplicate the `getProjectDetails()` call here and instantiate a second `Project` object.  It'd be easy to only call it once and store it in a variable that we reuse.

classes/ExternalModules.php
```diff
@@ -5997,36 +5997,40 @@ public static function isValidProjectId($pid, $condition = false) {
...
-		if ($row !== null) {
-			switch ($row["status"]) {
-				case 0: $status = "DEV"; break;
-				case 1: $status = "PROD"; break;
-				case 2: $status = empty($row["completed_time"]) ? "AC" : "DONE"; break;
-			}
-		}
-		return $status;
-	}
+    function getProjectDetails($pid){
```
mmcev106, line 6000: I think this method should be static.  It's referenced statically in the `Project` class below.

classes/ExternalModules.php
```diff
@@ -5997,36 +5997,40 @@ public static function isValidProjectId($pid, $condition = false) {
...
+        if (!self::isValidProjectId($pid)) {
+            throw new InvalidArgumentException(self::tt("em_errors_131")); //= Invalid value for project id!
+        }
```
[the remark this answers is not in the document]
knil-maloon: If you mean removed from getProjectStatus I think it's fine as getProjectDetails will do that check now.

classes/framework/Project.php
```diff
@@ -3,28 +3,40 @@
...
+    function getProjectId() {
+        return $this->project_id;
+    }
 
-	function getProjectId() {
-		return $this->project_id;
-	}
+    function getProjectTitle(){
+    return $this->getProjectDetails()->project['app_title'];
+    }
```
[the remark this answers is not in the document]
knil-maloon: makes sense

classes/ExternalModules.php
knil-maloon: changing it back

classes/ExternalModules.php
knil-maloon: done!

classes/ExternalModules.php
```diff
@@ -6023,11 +6023,11 @@ public static function getProjectStatus($pid){
...
             throw new InvalidArgumentException(self::tt("em_errors_131")); //= Invalid value for project id!
         }
-        $status = null;
-        switch (self::getProjectDetails($pid)->project['status']) {
+        $status = self::getProjectDetails($pid)->project['status'];
+        switch ($status) {
             case 0: $status = "DEV"; break;
             case 1: $status = "PROD"; break;
-            case 2: $status = empty(self::getProjectDetails($pid)->project["completed_time"]) ? "AC" : "DONE"; break;
+            case 2: $status = empty($status) ? "AC" : "DONE"; break;
```
mmcev106: This was checking `completed_time` previously in addition to `status`.  I think it might make sense to set the `project` to a variable instead of the `status`, so that we also have access to `completed_time`.

This method is unit tested now in the `testing` branch.  I would recommend running the test prior to committing additional changes.

knil-maloon: oops! I added the changes and made a test on my local. Seems to be working

classes/framework/Project.php
mmcev106: That is looking good to me.  Resolve...

classes/framework/Project.php
mmcev106: The new code is looking good.  See later comments for details.  Resolving...

classes/ExternalModules.php
mmcev106: The latest changes are looking good here.  Resolving...

classes/ExternalModules.php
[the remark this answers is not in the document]
mmcev106: I see it's removed now.  Looks good to me.  Resolving...

classes/framework/Project.php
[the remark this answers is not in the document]
mmcev106: I see you already resolved this.  If you don't mind, would it be OK if I resolved threads that I've opened, so I can use them as a checklist for reviewing each change?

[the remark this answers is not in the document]
mmcev106: This change does look good to me.

docs/framework/intro.md
```diff
@@ -119,6 +119,7 @@ The following methods are avaiable on the `Project` object returned by `$module-
 Method<br><br>&nbsp; | Minimum<br>REDCap<br>Version | Description<br><br>&nbsp;
 --- | --- | ---
 getProjectId() | ?.?.? | Returns the project id.
+getProjectTitle() | ?.?.? | Returns the project title.
```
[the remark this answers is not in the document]
mmcev106: The latest changes look good.

classes/ExternalModules.php
mmcev106: The latest changes look good.

classes/ExternalModules.php
mmcev106: The latest changes look good.

classes/ExternalModules.php
mmcev106: The latest changes look good.

classes/framework/Project.php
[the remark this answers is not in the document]
knil-maloon: Sounds good I thought as these changes were so straight forward that it would be ok to resolve. I will have it on mind.

classes/ExternalModules.php
knil-maloon: I've seen that now the throw argument comment does not make sense in there . Should I take that line out?

mmcev106: Hmm, what about it doesn't make sense?

knil-maloon: the @throws InvalidArgumentException as now it's in getDetails

mmcev106: Oh, I see what you mean.  I was looking at the "Invalid value for project" comment.  If you feel like moving that to the new method that would be great.  I guess I've trained myself to ignore those docblocks because I gave up on asking everyone to keep them up to date during PRs.

knil-maloon: done!

classes/ExternalModules.php
```diff
@@ -6016,7 +6016,6 @@ public static function getProjectDetails($pid){
      *
      * @param int|string $pid The project id.
      * @return null|string The status of the project. If the project does not exist, NULL is returned.
-     * @throws InvalidArgumentException
```
mmcev106: I'm curious, why was this removed rather than moving it to the new method where the `throw` was moved?

knil-maloon: Added a new comment with the throw and all I think it will be better also to have the function definition as well for getProjectDetails

mmcev106: Thank you!  Resolving...

PR #351, opened 2020-07-30
Add or update instances

mmcev106:
If y'all are able, I've love to get more eyes on this, especially to look for any potential long term support issues.  This "instance updated based on unique repeating field" pattern is once I realized I've reimplemented a few times in different projects, so here's my attempt at an abstracted version.


mmcev106:
I'm sorry for all the commit emails...  I thought I was done making changes when I created this PR.  Hopefully there aren't many more to follows.

mmcev106:
Much appreciate Kyle, thank you.  I just happened to notice that those functions were no longer used anywhere (and likely shouldn't have been added in the first place) so I went ahead and removed them since this touched related unit testing anyway.

mmcev106:
I think I'll go ahead and merge this with just Kyle's approval.  I copied the implementation into the Appointment Sync module and it seems to be working as expected.

PR #352, opened 2020-07-31
New Data Dictionary to Array function

knil-maloon:
Replaced function and added more options.


docs/framework/intro.md
```diff
@@ -47,6 +47,7 @@ createDAG($name) | ?.?.? | Creates a DAG with the specified name, and returns it
 createPassthruForm(<br>&emsp;$projectId,<br>&emsp;$recordId,<br>&emsp;<br>&emsp;// optional<br>&emsp;$surveyFormName,<br>&emsp;$eventId<br>) | ?.?.? | Outputs the HTML for opening/continuing the survey submission for the specified record.  If a return code is required, a link is returned.  Otherwise, a self submitting form is returned.
 createProject($title, $purpose, [, $project_note]) | 9.7.6 | Creates a new redcap project and returns the project id.
 createQuery() | 9.7.8 | Creates a `Query` object to aid in building complex queries using parameters.  See the [Query Documentation](../querying.md) page for more details.
+dataDictionaryExcelToArray($excelfilepath [, $returntype ]) | ?.?.? | Given a data dictionary file path, returns that file into an array.
```
mmcev106, line 50: Should we add supported options for the return type parameter to the description of this method?

mmcev106, line 50: I just standardized parameter names to use camel case.  Would you mind updating this documentation to match that standard?  I don't think we have to update it in the actual code.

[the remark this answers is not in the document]
knil-maloon: I think you're right it makes more sense to add CSV

knil-maloon: if they don't add anything by default it would be the A,B, C... type of array but maybe it should be good to add the options so they know what they can get?

mmcev106: As-is, if I was reading the docs I would think "Hmm, what are the options for that parameter and what do they do?".

knil-maloon: Done, changed the variable name as well accordingly to the first comment

knil-maloon: yup. My only thought is how to explain the default option of A, B, C.. csv array type?

knil-maloon: also would this way to display it work? $returnType={array, json}

mmcev106: I was thinking of it including it in the english description of the method.  Perhaps something like "The `$returnType` parameter controls {blank}.  Specify a value of {blank} to do {blank}".  Now that I'm thinking about it, does specifying `json` do anything different than simply calling `json_encode()` on the array output?  If it's not significantly different, we may want to rethink why we're supporting two different output formats.

mmcev106: Thanks!  Resolving...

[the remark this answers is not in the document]
mmcev106: Thanks for the update!  Resolving...

knil-maloon: it adds a nicer printing going record by record and adding a break the way REDCap does. If you think it is not necessary I can take it out.

mmcev106: Eva & I ended up having an offline conversation about options, and changed the way the method works a bit to simplify the publicly supported interface.  Resolving...

PR #357, opened 2020-08-14
New type on EM fpr yes/no

knil-maloon:
A new type so it only returns that yes/no type of field


mmcev106:
Copy of a slack chat between Eva & I:


Eva Bascompte Moragas:female_zombie:  11:40 AM
Created a new pull request for a new module I'm making. Something silly but I think very helpful user-wise





11:40
at least in my case






Mark McEver:bear:  11:40 AM
I just just looking at it.  Interesting...





11:41
I really like the idea of allowing for more control over dropdown field options, but only implementing this for yes-no fields seems restrictive.  I'm trying to think of a more generalized solution...






Eva Bascompte Moragas:female_zombie:  11:41 AM
I needed them to select a yesno type but having the whole field list won't help them





11:41
I think we could easily add several more types in that fashion (edited) 






Mark McEver:bear:  11:43 AM
What if instead we added a new hook, perhaps redcap_config_field_options($fieldName, $options) where you could override the option list for any field with any arbitrary PHP code?





11:43
It could also be implemented as a method in the config:
{
    "key": "some-field",
    "name": "Some Field",
    "type": "dropdown",
    "choices": 'somePHPMethodName()'
}
This is kinda similar to the custom fields examples we already have (though those are really old). (edited) 






Eva Bascompte Moragas:female_zombie:  11:45 AM
mmm but then you mean you would still have to call the sql that will show the fields? I'm not sure how the hook will work






Mark McEver:bear:  11:47 AM
Every time the config dialog is displayed, the options for each field could run through the hook to allow people to modify/replace them as needed for any field.






Eva Bascompte Moragas:female_zombie:  11:48 AM
Gotcha, that could work.
Another solution I thought is having field-type-* and have a regex to grab that type and add it in the sql so it's more dynamic.
I like your solution as then you let them add more things than that but that also worries me haha

Mark McEver:bear:  11:50 AM
Partial solutions have really held us back in the past.  I like the idea a lot, but I think it would make sense to give developers full control, and allowing them to control options via PHP seems like the only way to do that.

Eva Bascompte Moragas:female_zombie:  11:51 AM
Kyle was suggesting to do it as a property of field-type

Mark McEver:bear:  11:51 AM
We should make sure Gunther hasn't already started on something like this...  I think he's vacationing for another couple of weeks.  He might already have something like this built into the config 2.0 interface.

Eva Bascompte Moragas:female_zombie:  11:51 AM
I was talking with him when I had the idea for this haha

Mark McEver:bear:  11:51 AM
Awesome!

mmcev106:
This triggered a discussion between Eva, Kyle, and I about implement a more generalized hook for changing the rendered settings for any setting type.  I'll go ahead and close this since I think a different PR will replace it.

